Issue #49, opened 2014-11-19
Convert Lambda to Method Group whenever it's possible


ElemarJR:
Replaces

````csharp
input.Where(item => Test(item));
````

with 

````csharp
input.Where(Test);
````

Being worked on by @sqdavid
This is a `Hidden` dianostic.
Category is `Style`.
Diagnostid id is `CC0020`

giggio:
Closed on #72.

PR #72, opened 2014-11-26
Convert Lambda to Method Group when possible #49

sqdavid:
Replace a simple lambda expression that only invokes a method that has a
compatible signature with the method itself


hpshelton:
"ConvertLambdaExpressionToMethodGroupAnalizer.cs" should instead be named "ConvertLambdaExpressionToMethodGroupAnalyzer.cs". Be sure to update the class name as well.

giggio:
@hpshelton agreed. @sqdavid, please update.
Also, @sqdavid, would you please squash the commits? I am not sure how well you know git, so just to help: you do it like that: `git rebase -i master`. Be careful because you may end up messing up the commits, so check everything before you `git push --force` it.

sqdavid:
updated the class name.

giggio:
@sqdavid it is merged. I removed the unnecessary check happening on the code fix provider. You were already checking that on the speculative model, you did not need to check it again. All tests are passing. If that code was necessary we need a test to show it. Look at ee728b6 to see the changes.

sqdavid:
Great! I think the removed check was not needed

PR #92, opened 2014-11-28
Analyzer to check if disposable types are calling GC.SuppressFinalize

israelaece:


andrecarlucci:
I know this is old, but... my class is not dealing with unmanaged code and the analyzer is telling me to mess with the GC.
Is this really good?

sharwell:
@andrecarlucci Unless your type is `sealed` *and* does not have a finalizer, your implementation of `IDisposable.Dispose()` should contain a call to `GC.SuppressFinalize(this)` as the last statement.

andrecarlucci:
My class has an internal HttpClient used in multiple requests and I implement IDisposable to dispose it when I'm finished. What problem GC.SupressFinalize is solving?

sharwell:
@andrecarlucci If a class extends that internal HttpClient and adds a finalizer, it can override the protected `Dispose(bool)` method to release resources but it can't add the `GC.SuppressFinalize()` call to `Dispose()`.

Since your class is internal, you don't have to worry about third parties extending it like this. The specific problem the call is solving is a *maintainability* problem. It keeps you from having to worry about whether or not a base class is following rules if a finalizer is ever used. As mentioned earlier, the other alternative is marking the class `sealed`.

:memo: Note that the call to `GC.SuppressFinalize` is *almost* free when the type does not have a user-defined finalizer.

andrecarlucci:
But that case is an anti-pattern itself.
If you want to add a finalizer to deal with unmanaged resources, that class should not extend any other class. It should be as simple as it can be and only deal with the unmanaged resource.
That way child classes would never have to deal with the GC.
It's not a question of performance hit, it's a question about adding an extra line that developers have to look and think why it's there if it's doing nothing.

sharwell:
It's like a seat belt. Most of the time it isn't doing anything. When it is there's a good chance it's because someone did something wrong, but in that time it does its job.

andrecarlucci:
I got you, thanks for clarifying your thoughts.
I personally don't think it's worth it for my scenarios (enterprise), but I can see how things can go wrong with open-source projects where you don't have control about who's using them. Cheers. :)

PR #183, opened 2015-01-02
IDisposable variable not assigned to a field is not being disposed

giggio:
I dealed with several interesting cases, like explicit implementation of IDisposable, and had some fun with data flow analysis as well. Read the tests for better understanding.

This is the end result:

image

Docs are updated: http://code-cracker.github.io/diagnostics/CC0022.html


giggio:
Hold it, I forgot to ignore returns.

PR #184, opened 2015-01-02
Allow members ordering by style cop patterns (Closes #172)

### LINKED ISSUE #172: Offer a fix for ordering members inside classes and structs following StyleCop patterns
Regarding the issue [Offer diagnostic to allow for ordering members inside classes and structs](https://huboard.com/code-cracker/code-cracker#/issues/50224648). 
We could offer a fix for ordering classes and structs members following [StyleCop documentation](http://www.stylecop.com/docs/Ordering%20Rules.html) as recommended by @paulomorgado .

1. [Within a class, struct, or interface, elements must be positioned in the following order:](http://www.stylecop.com/docs/SA1201.html)
 * Fields
  * Constructors
  * Finalizers (Destructors)
  * Delegates
  * Events
  * Enums
  * Interfaces
  * Properties
  * Indexers
  * Methods
  * Structs
  * Classes

2. [Adjacent elements of the same type must be positioned in the following order by access level:](http://www.stylecop.com/docs/SA1202.html)
  * public
  * internal
  * protected internal
  * protected
  * private

3. Fields Constant [A constant field is placed beneath a non-constant field](http://www.stylecop.com/docs/SA1203.html)

4. [All static elements must be placed above all instance elements of the same type](http://www.stylecop.com/docs/SA1204.html)

5. And finally by alphabetical order. This is my personal suggestion and is **not** part of StyleCop documentation. 

This source:

```C#
            using System;
            namespace ConsoleApplication1
            {
                class Foo
                {
                    public class Foo2 { }
                    public struct Struct { }
                    public static Foo operator +(Foo f1, Foo f2) { return new Foo(); }
                    public static Foo operator -(Foo f1, Foo f2) { return new Foo(); }
                    void Method(string a) { }
                    internal void Method(int a) { }
                    public void Method1() { }
                    public void Method() { }
                    public string this[int i] { set { } }
                    public string Property { get; set; }
                    public interface Interface { }
                    public enum Enum { Enum1, Enum2 = 1 }
                    public event Action Event 
                    { 
                        add { EventField += value; } 
                        remove { EventField -= value; } 
                    }
                    public delegate double Delegate(double num);
                    public Foo()
                    {
                        Property = Field1 = Field = "";
                        EventField = EventField1 = () => { };
                    }
                    public event Action EventField1;
                    public event Action EventField;
                    public string Field;
                    public static string Field1;
                }
            }
```

Became:
```C#
            using System;
            namespace ConsoleApplication1
            {
                class Foo
                {
                    public static string Field1;
                    public string Field;
                    public Foo()
                    {
                        Property = Field1 = Field = "";
                        EventField = EventField1 = () => { };
                    }
                    public delegate double Delegate(double num);
                    public event Action EventField;
                    public event Action EventField1;
                    public event Action Event 
                    { 
                        add { EventField += value; } 
                        remove { EventField -= value; } 
                    }
                    public enum Enum { Enum1, Enum2 = 1 }
                    public interface Interface { }
                    public string Property { get; set; }
                    public string this[int i] { set { } }
                    public static Foo operator -(Foo f1, Foo f2) { return new Foo(); }
                    public static Foo operator +(Foo f1, Foo f2) { return new Foo(); }
                    public void Method() { }
                    public void Method1() { }
                    internal void Method(int a) { }
                    void Method(string a) { }
                    public struct Struct { }
                    public class Foo2 { }
                }
            }
```

@dmgandini is working on it.

dmgandini:
https://github.com/code-cracker/code-cracker/issues/172

I created the BaseAllowMembersOrderingCodeFixProvider in order to increase code reuse as more ordering algorithms are expected in the future. 

Each <NameX>AllowMembersOrderingCodeFixProvider should provides a IComparer in order to put the type members on the correct order and the BaseAllowMembersOrderingCodeFixProvider takes the responsibility to write them.

### DIFF
--- a/src/CSharp/CodeCracker/Refactoring/AllowMembersOrderingCodeFixProvider.Base.cs
+++ b/src/CSharp/CodeCracker/Refactoring/AllowMembersOrderingCodeFixProvider.Base.cs
@@ -32,11 +32,9 @@ namespace CodeCracker.Refactoring
                         .FindToken(diagnosticSpan.Start)
                         .Parent as TypeDeclarationSyntax;
 
-            context.RegisterFix(
-                CodeAction.Create(
-                    codeActionDescription,
-                    cancellationToken => AllowMembersOrderingAsync(context.Document, typeDeclarationSyntax, cancellationToken)),
-                    diagnostic);
+            var newDocument = await AllowMembersOrderingAsync(context.Document, typeDeclarationSyntax, context.CancellationToken);
+            if (newDocument != null)
+                context.RegisterFix(CodeAction.Create(codeActionDescription, newDocument), diagnostic);
         }
 
         private async Task<Document> AllowMembersOrderingAsync(Document document, TypeDeclarationSyntax typeDeclarationSyntax, CancellationToken cancellationToken)
@@ -48,10 +46,14 @@ namespace CodeCracker.Refactoring
 
             var root = await document.GetSyntaxRootAsync(cancellationToken) as CompilationUnitSyntax;
 
-            var newTypeDeclarationSyntax = ReplaceTypeMembers(
+            TypeDeclarationSyntax newTypeDeclarationSyntax;
+            var orderChanged = TryReplaceTypeMembers(
                 typeDeclarationSyntax,
                 membersDeclaration,
-                membersDeclaration.OrderBy(member => member, GetMemberDeclarationComparer(document, cancellationToken)));
+                membersDeclaration.OrderBy(member => member, GetMemberDeclarationComparer(document, cancellationToken)),
+                out newTypeDeclarationSyntax);
+
+            if (!orderChanged) return null;
 
             var newDocument = document.WithSyntaxRoot(root
                  .ReplaceNode(typeDeclarationSyntax, newTypeDeclarationSyntax)
@@ -63,13 +65,20 @@ namespace CodeCracker.Refactoring
 
         protected abstract IComparer<MemberDeclarationSyntax> GetMemberDeclarationComparer(Document document, CancellationToken cancellationToken);
 
-        private TypeDeclarationSyntax ReplaceTypeMembers(TypeDeclarationSyntax typeDeclarationSyntax, IEnumerable<MemberDeclarationSyntax> membersDeclaration, IEnumerable<MemberDeclarationSyntax> sortedMembers)
+        private bool TryReplaceTypeMembers(TypeDeclarationSyntax typeDeclarationSyntax, IEnumerable<MemberDeclarationSyntax> membersDeclaration, IEnumerable<MemberDeclarationSyntax> sortedMembers, out TypeDeclarationSyntax orderedType)
         {
             var sortedMembersQueue = new Queue<MemberDeclarationSyntax>(sortedMembers);
+            var orderChanged = false;
 
-            return typeDeclarationSyntax.ReplaceNodes(
+            orderedType = typeDeclarationSyntax.ReplaceNodes(
                 membersDeclaration,
-                (original, rewritten) => sortedMembersQueue.Dequeue());
+                (original, rewritten) =>
+                {
+                    var newMember = sortedMembersQueue.Dequeue();
+                    if (!orderChanged && !original.Equals(newMember)) orderChanged = true;
+                    return newMember;
+                });
+            return orderChanged;
         }
 
         public override ImmutableArray<string> GetFixableDiagnosticIds()
--- a/test/CSharp/CodeCracker.Test/Refactoring/AllowMembersOrderingCodeFixProviderTests.Base.cs
+++ b/test/CSharp/CodeCracker.Test/Refactoring/AllowMembersOrderingCodeFixProviderTests.Base.cs
@@ -37,6 +37,19 @@ namespace CodeCracker.Test.Refactoring
             Assert.True(codeFixProvider.HasIComparerBeenCalled, "The IComparer must be used to sort the members of that type");
         }
 
+        [Fact]
+        public async Task BaseAllowMembersOrderingShouldNotRegisterFixIfIsAlreadySorted()
+        {
+            var source = @"
+            public class Foo
+            {
+                int a;
+                void B() { }
+            }";
+
+            await VerifyCSharpHasNoFixAsync(source);
+        }
+
         [Theory]
         [InlineData("class")]
         [InlineData("struct")]
--- a/test/CSharp/CodeCracker.Test/Refactoring/AllowMembersOrderingCodeFixProviderTests.StyleCop.cs
+++ b/test/CSharp/CodeCracker.Test/Refactoring/AllowMembersOrderingCodeFixProviderTests.StyleCop.cs
@@ -30,11 +30,11 @@ namespace CodeCracker.Test.Refactoring
             const string expected = @"
             class Foo
             {
-                public int b = 0;
-                public event System.Action EventField;
+                public int b = 0;                
                 Foo() { }
                 ~Foo() { }
                 public delegate double Delegate();
+                public event System.Action EventField;
                 public event System.Action Event { add { } remove { } }
                 enum Enum { Enum1 }
                 public interface Interface { }
@@ -69,13 +69,13 @@ namespace CodeCracker.Test.Refactoring
             public class Foo
             {
                 public const int z = 0;
-                const int x = 0;
                 public static int v = 0;
-                static int u = 0;
                 public int t = 0;
                 internal int s = 0;
                 protected internal int r = 0;
                 protected int q = 0;
+                const int x = 0;
+                static int u = 0;
                 private int p = 0;
             }";
 
@@ -116,11 +116,11 @@ namespace CodeCracker.Test.Refactoring
             class Foo
             {
                 int a = 0, b = 1;
-                int c = 2, d = 3;
-                event System.Action EventField;
-                event System.Action EventField1;
+                int c = 2, d = 3;        
                 delegate double Delegate();
                 delegate double Delegate2();
+                event System.Action EventField;
+                event System.Action EventField1;
                 event System.Action Event { add { } remove { } }
                 event System.Action Event1 { add { } remove { } }
                 enum Enum { Enum1 }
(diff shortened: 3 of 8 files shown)


src/CSharp/CodeCracker/CodeCracker.csproj
```diff
@@ -36,6 +36,8 @@
     <Compile Include="Performance\StringBuiderInLoopCodeFixProvider.cs" />
     <Compile Include="Performance\StringBuilderInLoopAnalyzer.cs" />
     <Compile Include="Refactoring\AllowMembersOrderingAnalyzer.cs" />
+    <Compile Include="Refactoring\AllowMembersOrderingCodeFixProvider.StyleCop.cs" />
+    <Compile Include="Refactoring\AllowMembersOrderingCodeFixProvider.Base.cs" />
```
dmgandini: Instead of create the file with the same name of the class, I chose to use the AllowMembersOrderingCodeFixProvider as a prefix to make easy to locate each implementation of BaseAllowMembersOrderingCodeFixProvider

src/CSharp/CodeCracker/Refactoring/AllowMembersOrderingCodeFixProvider.Base.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            var newDocument = document.WithSyntaxRoot(root
+                 .ReplaceNode(typeDeclarationSyntax, newTypeDeclarationSyntax)
+                 .WithAdditionalAnnotations(Formatter.Annotation)
+            );
+
+            return newDocument;
+        }
+
+        protected abstract IComparer<MemberDeclarationSyntax> GetMemberDeclarationComparer(Document document, CancellationToken cancellationToken);
```
dmgandini: I'm passing the Document and the CancellationToken on that method, because a semantic model can be needed in a derived class.

test/CSharp/CodeCracker.Test/Refactoring/AllowMembersOrderingCodeFixProviderTests.Base.cs
```diff
@@ -0,0 +1,137 @@
...
+
+            var expected = @"
+            public " + typeDeclaration + @" Foo
+            {
+                " + memberB + @"
+                " + memberA + @" 
+            }";
+
+            await VerifyCSharpFixAsync(source, expected, codeFixProvider: codeFixProvider);
+            Assert.True(codeFixProvider.HasIComparerBeenCalled, "The IComparer must be used to sort the members of that type");
```
dmgandini: I could use just VerifyCSharpAsync as a proof of IComparer was being called, but for readability sake I used a mock.

test/CSharp/CodeCracker.Test/Verifiers/CodeFixVerifier.cs
```diff
@@ -42,14 +42,15 @@ protected virtual CodeFixProvider GetBasicCodeFixProvider()
         /// <param name="newSource">A class in the form of a string after the CodeFix was applied to it</param>
         /// <param name="codeFixIndex">Index determining which codefix to apply if there are multiple</param>
         /// <param name="allowNewCompilerDiagnostics">A bool controlling whether or not the test will fail if the CodeFix introduces other warnings after being applied</param>
-        protected async Task VerifyCSharpFixAsync(string oldSource, string newSource, int? codeFixIndex = null, bool allowNewCompilerDiagnostics = false, bool formatBeforeCompare = true)
+        /// <param name="codeFixProvider">The codefix to be applied to the code wherever the relevant Diagnostic is found</param>
+        protected async Task VerifyCSharpFixAsync(string oldSource, string newSource, int? codeFixIndex = null, bool allowNewCompilerDiagnostics = false, bool formatBeforeCompare = true, CodeFixProvider codeFixProvider = null)
```
dmgandini: The added codeFixProvider parameter made possible to pass a mock object to VerifyCSharpFix Async method.
I couldn't override the CSharpCodeProvider method, because the multithread nature of our tests.

src/CSharp/CodeCracker/CodeCracker.csproj
giggio: I like it.

test/CSharp/CodeCracker.Test/Verifiers/CodeFixVerifier.cs
giggio: Why not just create another test class? I think it would be easier and wouldn't require changing this method. Right?

giggio: Never mind, I understand what you are trying to do there.

src/CSharp/CodeCracker/Refactoring/AllowMembersOrderingCodeFixProvider.Base.cs
giggio: Ok, understood.

giggio:
Build is passing: https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.221

giggio:
@dmgandini I got an error when running in VS and the class was already sorted.
I don't think we should present a fix if it is already sorted. How can we fix this?

giggio:
Ok, I was able to add a quick fix. I will commit it now and move on.

test/CSharp/CodeCracker.Test/Refactoring/AllowMembersOrderingCodeFixProviderTests.StyleCop.cs
```diff
@@ -0,0 +1,225 @@
...
+                public const int z = 0;
+                const int x = 0;
+                public static int v = 0;
+                static int u = 0;
+                public int t = 0;
+                internal int s = 0;
+                protected internal int r = 0;
+                protected int q = 0;
+                private int p = 0;
+            }";
```
giggio: This order seems incorrect.
`const int x = 0` is the same as `private const int x = 0`. 
Also, on #172 we remember that we need to have constant fields bellow not constant fields.
So, shouldn't we have all public fields before all private fields?
And static before instance?

Like this:

````csharp
            public class Foo
            {
                static int u = 0;
                public const int z = 0;
                public static int v = 0;
                public int t = 0;
                internal int s = 0;
                protected internal int r = 0;
                protected int q = 0;
                private int p = 0;
                const int x = 0;
            }
````

**Comment by giggio** [comment]
@dmgandini please review the ordering question above. Also, I updated my branch with the unnecessary ordering fix on aac6d0c, check it out, please.

File: test/CSharp/CodeCracker.Test/Refactoring/AllowMembersOrderingCodeFixProviderTests.StyleCop.cs (hunk 5 above)
Comment (dmgandini) [inline test/CSharp/CodeCracker.Test/Refactoring/AllowMembersOrderingCodeFixProviderTests.StyleCop.cs]: I believe the test is correct,  I applied the rules in the following order:
1. [Constants Must Appear Before Fields] (http://www.stylecop.com/docs/SA1203.html)
2. [All static elements must be placed above all instance elements of the same type] (http://www.stylecop.com/docs/SA1204.html)
3. [Elements Must Be Ordered By Access] (http://www.stylecop.com/docs/SA1202.html)
4. Alphabetical order

When a member has no access keyword, such as private/public, the private value is assumed, thus ```const int x = 0``` is the same as ```private const int x = 0``` as you said.
```const int x = 0;``` is the second to appear because is a private const. So it came after a public const field.

Do you agree?

**Comment by dmgandini** [comment]
@giggio I got your commit and added to this PR. I also did create a test regarding to the fix.

File: test/CSharp/CodeCracker.Test/Refactoring/AllowMembersOrderingCodeFixProviderTests.StyleCop.cs (hunk 5 above)
Comment (giggio) [inline test/CSharp/CodeCracker.Test/Refactoring/AllowMembersOrderingCodeFixProviderTests.StyleCop.cs]: I am not sure. Which rule has precedence? Do we know that?

Comment (dmgandini) [inline test/CSharp/CodeCracker.Test/Refactoring/AllowMembersOrderingCodeFixProviderTests.StyleCop.cs]: Actually, I didn't find any document where the precedence is specified.
So, I've used the rules descriptions to try to figure it out.

I'm not sure what come first: const or static when the members are fields. I reckon  "ConstantsMustAppearBeforeFields" is quite more imperative than "StaticElementsMustAppearBeforeInstanceElements", also static fields are still fields.

Comment (giggio) [inline test/CSharp/CodeCracker.Test/Refactoring/AllowMembersOrderingCodeFixProviderTests.StyleCop.cs]: Ok, I am checking this with Microsoft.

**Comment by giggio** [comment]
To anyone following this PR: I am verifying the StyleCop ordering with Microsoft. If anyone has any docs on this, please point them out so we can look into them.

File: test/CSharp/CodeCracker.Test/Refactoring/AllowMembersOrderingCodeFixProviderTests.StyleCop.cs (hunk 5 above)
Comment (dmgandini) [inline test/CSharp/CodeCracker.Test/Refactoring/AllowMembersOrderingCodeFixProviderTests.StyleCop.cs]: Thanks

**Comment by dmgandini** [comment]
@giggio I've installed the package StyleCop.MSBuild, version 4.7.49.0, into a test project. I found out that some tests are incorrect. 
First I will update the tests description and after it will be applied to this PR.
Please, do

File: test/CSharp/CodeCracker.Test/Refactoring/AllowMembersOrderingCodeFixProviderTests.StyleCop.cs (hunk 5 above)
Comment (dmgandini) [inline test/CSharp/CodeCracker.Test/Refactoring/AllowMembersOrderingCodeFixProviderTests.StyleCop.cs]: The correct order according to StyleCop.MSBuild, version 4.7.49.0:  

```c#
    public class Foo
    {
        public const int z = 0;
        public static int v = 0;        
        public int t = 0;
        internal int s = 0;
        protected internal int r = 0;
        protected int q = 0;
        const int x = 0;
        static int u = 0;
        private int p = 0;
    }
```

1. Elements Must Be Ordered By Access
2. Constants Must Appear Before Fields
3. All static elements must be placed above all instance elements of the same type

giggio:
Good job! I will wait for the updated code.

giggio:
@dmgandini I saw you pushed new code. Are you done?

dmgandini:
I think it is done.

giggio:
Build is passing: https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.295

giggio:
@dmgandini I made a little change on the code fix message so it shows using the type name, like `Order {0}'s members following StyleCop patterns`. Good job!

dmgandini:
Perfect, thanks man!

PR #185, opened 2015-01-02
Const fields can't be made readonly

felipebz:
Fixes CC0052 (Make field readonly) to ignore const fields.


carloscds:
@felipebz Thanks for fix!

PR #186, opened 2015-01-02
Fixes NullReferenceException in CC0047

felipebz:
Expression bodied properties don't have an AccessorList, so we need a null check.

### DIFF
--- a/src/CSharp/CodeCracker/Style/PropertyPrivateSetAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Style/PropertyPrivateSetAnalyzer.cs
@@ -39,7 +39,7 @@ namespace CodeCracker.Style
             var invocationExpression = (PropertyDeclarationSyntax)context.Node;
             var semanticModel = context.SemanticModel;
 
-            if (invocationExpression.AccessorList.Accessors.Count == 1) return;
+            if (invocationExpression.AccessorList == null || invocationExpression.AccessorList.Accessors.Count == 1) return;
 
             var setAcessor = (invocationExpression.AccessorList.Accessors[0].Keyword.Text == "set") ? invocationExpression.AccessorList.Accessors[0] : invocationExpression.AccessorList.Accessors[1];
 
--- a/test/CSharp/CodeCracker.Test/Style/PropertyPrivateSetTests.cs
+++ b/test/CSharp/CodeCracker.Test/Style/PropertyPrivateSetTests.cs
@@ -46,6 +46,22 @@ namespace CodeCracker.Test.Style
         }
 
         [Fact]
+        public async Task ExpressionBodiedPropertyDoesNotCreateDiagnostic()
+        {
+            const string test = @"
+    using System;
+
+    namespace ConsoleApplication1
+    {
+        class TypeName
+        {
+            public int MyProperty => 0;
+        }
+    }";
+            await VerifyCSharpHasNoDiagnosticsAsync(test);
+        }
+
+        [Fact]
         public async Task PropertyPrivateFixAutoProperty()
         {
             const string test = @"


src/CSharp/CodeCracker/Style/PropertyPrivateSetAnalyzer.cs
```diff
@@ -39,7 +39,7 @@ private static void AnalyzeClass(SyntaxNodeAnalysisContext context)
             var invocationExpression = (PropertyDeclarationSyntax)context.Node;
             var semanticModel = context.SemanticModel;
 
-            if (invocationExpression.AccessorList.Accessors.Count == 1) return;
+            if (invocationExpression.AccessorList == null || invocationExpression.AccessorList.Accessors.Count == 1) return;
```
viniciushana: Perhaps this could be `invocationExpression.AccessorList?` instead?

felipebz: Sure, but the if would have to change to:

    (invocationExpression.AccessorList?.Accessors.Count ?? 1) == 1

Should I change it?

giggio: I find it more readable, but it is a detail...

carloscds: Note: This analyzer works with Properties, and even so, if there aren´t acessors, it not a property, it´s a Field and analyzer will not called. It´s not necessary check AcessorList == null. For a property, always has almost one acessor: get or set.

felipebz: @carloscds please, see my test case. In C# 6 we have expression bodied properties, like `public int MyProperty => 0;`. They are getter-ony properties and don't have a get acessor.

carloscds: @felipebz Ok, It's make sense. In C# 6 you can do this.

paulomorgado: I think taht, in this case, you can even shorten it more:

    invocationExpression.AccessorList?.Accessors.Count == 1

If `invocationExpression.AccessorList` is `null`, `invocationExpression.AccessorList?.Accessors.Count` won't be `1`.

I don't have VS2015 here, but I tried this:

    using System;
    using System.Collections.Generic;
    public class C
    {
        public void M()
        {
            Dictionary<string,string> d = null;
            if (d?.Keys.Count == 1)
            {
            }
        }
    }

in [http://tryroslyn.azurewebsites.net/](http://tryroslyn.azurewebsites.net/#K4Zwlgdg5gBAygTxAFwKYFsDcAoUlaIoYB0AwgPYA2lqAxsmORCMQOKoSoBOYtOADsABGlXjFqUAhiBAxS2AN7YYKmIJFiAbuTAATGAFkAFAEplqpaqswAIrwZNJXBAB4UPaABp3+AHwx9AF4YCGBqHGtVMAAzGCNdAH5iAGlUJDJyYAhkGEDggEYzSJVLYoBfcxUKsqAAA=)

felipebz: @paulomorgado When `invocationExpression.AccessorList` is null:

`invocationExpression.AccessorList == null || invocationExpression.AccessorList.Accessors.Count == 1` returns `true`

`invocationExpression.AccessorList?.Accessors.Count == 1`returns `false`. That is wrong.

paulomorgado: You're right. Reading code on the phone and reason about it doesn't seem to be a good idea. :smile:

But I'll try again.

.    invocationExpression.AccessorList?.Accessors.Count.GetValueOrDefault(1) == 1

felipebz: Right, this works: `(invocationExpression.AccessorList?.Accessors.Count).GetValueOrDefault(1) == 1`

but it is equals to `(invocationExpression.AccessorList?.Accessors.Count ?? 1) == 1` as discussed above :smiley:

Issue #100, opened 2014-12-01
Run Analysis on a large OSS project and make sure it does not throw


giggio:
Sometimes our analyzers throw exceptions when used on real code. The tests alone are not able to make sure they are not breaking, as they are limited by the programmers perspective. We should be able to compile a large open source project and make sure the analyzers don't throw.
The idea is to have a simple command line tool that we can run that will clone some project, run the analyzers and fail if something throws. And this tool will be wired during the build so it also works on every push and PR.

giggio:
I will start working on this issue. The idea is to start using https://github.com/dotnet/corefx. I will clone, update all .csproj files to use our analyzers and run the builds and make sure it does not throw anywhere. Should be a simple xml transformation, followed by msbuild on each csproj and check on errorlevel.

giggio:
Ok, corefx is done. And I have just found some bugs. :) See the last lines of the build:
https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.241

giggio:
Now I need projects suggestions to work on. I have tried Umbraco and Xunit and they are both too complex to bring into this process. @code-cracker/owners need your help.

giggio:
We have now analyzis on Corefx and MonoCecil. It is good for now, and it will run on every build. It is not breaking the build yet because of bug #192 that @ElemarJR is fixing. Once that is done I will change the build script and if a problem is found the build will fail.

PR #187, opened 2015-01-04
Analyze CoreFX after tests

giggio:
Issue #100
Updated appveyor build file.

Note:
After analyzing figured that CC0004 (EmptyCatchBlockAnalyzer)
was using Error as severity, and this was not allowing corefx
to build, so we figured it was a common pattern, so it is now
a warning.

### DIFF
--- a/CodeCracker.CSharp.sln
+++ b/CodeCracker.CSharp.sln
@@ -19,9 +19,15 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
 	ProjectSection(SolutionItems) = preProject
 		.gitattributes = .gitattributes
 		.gitignore = .gitignore
+		appveyor.yml = appveyor.yml
 		buildNuget.cmd = buildNuget.cmd
 		src\CSharp\buildNugetCSharp.cmd = src\CSharp\buildNugetCSharp.cmd
 		README.md = README.md
+	EndProjectSection
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "test", "test", "{051F1BE2-9A44-4B84-9DF8-6537852B7BBC}"
+	ProjectSection(SolutionItems) = preProject
+		test\CSharp\AnalyzeCoreFx.ps1 = test\CSharp\AnalyzeCoreFx.ps1
 		runTestsCS.ps1 = runTestsCS.ps1
 	EndProjectSection
 EndProject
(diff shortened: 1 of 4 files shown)


giggio:
I usually don't do a PR for infrastructure stuff, but i want to see if this builds before. I am taking care of it myself.

paulomorgado:
Maybe we shouldn't rely that much on corefx complying with anything. Even their own rules.

See [this thread](http://forums.dotnetfoundation.org/t/underscores-in-private-fields/731 "Underscores in private fields") on the [](Maybe we shouldn't rely that much on corefx complying with anything. Even their own rules.

See [this thread](http://forums.dotnetfoundation.org/t/underscores-in-private-fields/731 "Underscores in private fields ") on the [.NET Foundation Forums](http://forums.dotnetfoundation.org/http://forums.dotnetfoundation.org/ ".NET Foundation Forums").

giggio:
@paulomorgado This work is not meant to check style. I am aware that a lot of diagnostics will show up. What I don't want is to for CodeCracker to throw while analyzing. I actually found 3 bugs yesterday because of this analysis. See the last lines of the last build for example, in red:
https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.247
You will notice some exceptions after analyzing corefx. That is the goal.
BTW, I agree with you on the styles discussions, I will chime in. :)

giggio:
@jwooley I am looking for a large open source VB project(s) to run CodeCracker against during build. We already do that with CoreFX and Cecil, but they are both C#. Do you know of any?
See what's happening here: https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.408#L281
Also the [AnalyzeCoreFx.ps1](https://github.com/code-cracker/code-cracker/blob/master/test/CSharp/AnalyzeCoreFx.ps1) file.

jwooley:
One option: https://github.com/dotnet/roslyn/tree/master/src/Compilers/VisualBasic/Portable



From the original LINQ samples (in 2008): https://code.msdn.microsoft.com/Official-Visual-Studio-f48134ec

The original link is now dead, but the EF Samples did have a fair amount of VB: http://web.archive.org/web/20110818102600/http://adonetsamples.codeplex.com/releases/view/8858

You’re fully welcome to try any of the samples at http://www.thinqlinq.com/Files.aspx as well




Sent from Windows Mail





From: Giovanni Bassi
Sent: ‎Monday‎, ‎February‎ ‎16‎, ‎2015 ‎9‎:‎49‎ ‎AM
To: 'code-cracker/code-cracker'
Cc: Jim Wooley





@jwooley I am looking for a large open source VB project(s) to run CodeCracker against during build. We already do that with CoreFX and Cecil, but they are both C#. Do you know of any?
See what's happening here: https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.408#L281
Also the AnalyzeCoreFx.ps1 file.

giggio:
@jwooley We can't do roslyn code until they freeze it. I tried with C# and it throws all sorts of errors.
For the others I need a public git repo that we can pull in during build.

Issue #190, opened 2015-01-04
Refactor Regex and JsonNet Analyzers to be similar to UriAnalyzer


andrecarlucci:
First of all, I hate this class name: "PreRunMethodChecker". I will change it as soon as I find a better one.

Since creating Analyzers that run a method during design time to check if it would throw at runtime is a lot of copy/pasting with some tweaks, I created a helper class to simplify adding new ones.

The idea is (UriAnalyzer example):

```
   private void Analyzer(SyntaxNodeAnalysisContext context)
   {
            var mainConstrutor = new PreRunMethodInfo(
                "System.Uri.Uri(string)",
                args => { new Uri(args[0].ToString()); }
            );
            var constructorWithUriKind = new PreRunMethodInfo(
                "System.Uri.Uri(string, System.UriKind)",
                args => { new Uri(args[0].ToString(), (UriKind)args[1]); }
            );

            var checker = new PreRunMethodChecker(context, Rule);
            checker.AnalyzeConstrutor(mainConstrutor);
            checker.AnalyzeConstrutor(constructorWithUriKind);
   }
```

It also has a method called "AnalyzeMethod" that works the same way as "AnalyzeConstrutor": you pass the method signature you want to find and a action that will be invoked inside a try/catch test once the code is found.

I thought about changing the Regex and JsonNet analyzers to use this too so we have code consistence, but it's not really necessary.

giggio:
@andrecarlucci You already did this, right?

giggio:
@andrecarlucci You already did this, right?

andrecarlucci:
Not really, sorry. I'll do it after the virtual one.

giggio:
@andrecarlucci Do you plan to come back to this? If not, can you elaborate on what was in your plans?

andrecarlucci:
I'll do it, just give me a week :)

Issue #182, opened 2014-12-30
Validade Uri from System.Uri


andrecarlucci:
This:

`
var x = new Uri("wrong uri");
`

Will be marked as an `Error`. 
Category is `Usage`.
Diagnostic id is `CC0063`

andrecarlucci:
Thanks, I edited my message and updated the wiki.
I will start working on this.

giggio:
Deal, it is yours.

PR #191, opened 2015-01-04
Validade uri from systemuri #182

andrecarlucci:
See https://github.com/code-cracker/code-cracker/issues/190 to understand the reason for PreRunMethodChecker and PreRunMethodInfo classes.


carloscds:
@andrecarlucci André, your analyzer is being executed for all object creation and not only for Uri. Can you check this ?

andrecarlucci:
Sorry, I thought It has supposed to do that. I mean, it has to run for every constructor in order to check if it's the desired constructor. Is there another way?

Or do mean that it is returning a diagnostic for other class constructor? Can you provide me a sample that would show me the error? I'll fix asap :)
Tks!

carloscds:
@andrecarlucci No problem, take a look ar RegexAnalizer. It´s same kind of validation.

andrecarlucci:
The RegexAnalizer is not validating a constructor, it uses the InvocationExpressionSyntax. For constructors we need ObjectCreationExpressionSyntax. The analyser is also called for every method, isn't it?

Also, it does not validade Regex with RegexOptions neither the Regex constructor. I want to later apply this refactory to fix that: https://github.com/code-cracker/code-cracker/issues/190, but that's another issue.

carloscds:
@andrecarlucci My suggestion is that you verify is object is Uri. ObjectCreationExpression is executed for all object creation, for example, if a create SqlConnection() this analyzer is called. InvocationExpress is only for expression, like parameters in Regex.

carloscds:
@andrecarlucci Using ObjectCreationExpression you can check IdentiferName looking for Uri

andrecarlucci:
Now I got you. There are two different things here:

The UriAnaylizer will run for every constructor found in the same way the RegexAnalizer runs for every invocation method found. RegexAnalizer runs for sqlConnection.Open() for instance, that's the way analyzers work.

The optimization you're looking for is to return as soon as the analyzer finds that the method name is wrong using only the syntax tree. Doing so avoids loading the heavy weight semantic model into memory unnecessary.

ObjectCreationExpressionSyntax doesn't have IdentiferName or a Identifier property the way MemberAccessExpressionSyntax does, so I can't follow the same path. 
Since this information is part of the syntax, it's definitely there somewhere. I'll find it, update the code and commit the changes.

Thanks a lot for pointing me that out.

carloscds:
@andrecarlucci Let me know when fix is done for commit.

andrecarlucci:
@carloscds There you go :)

carloscds:
@andrecarlucci Thanks for your PR André! Send us more :-)

carloscds:
@andrecarlucci André, please check exception messages in test, because in a diferente en-US system, exceptions returns in a different language, not always english and tests can fail.

andrecarlucci:
Good point.
So we don't test exception messages for these cases, right? Just compare whatever it returns?
Do I send a new PR referencing the same issue or is it better to add a new one?

viniciushana:
IMO I see no harm in referencing the same issue in another PR

carloscds:
@andrecarlucci Can you send using same issue. Only comment about what you are fixing/improving.

src/CSharp/CodeCracker/Usage/MethodAnalyzers/MethodChecker.cs
```diff
@@ -0,0 +1,106 @@
...
+            var objectCreationExpressionSyntax = (ObjectCreationExpressionSyntax) context.Node;
+            var identifier = objectCreationExpressionSyntax.Type as IdentifierNameSyntax;
+            return identifier?.Identifier.ValueText != methodInformation.MethodName;
+        }
+
+        private bool QualifiedConstructorNameNotFound(MethodInformation methodInformation)
+        {
+            var objectCreationExpressionSyntax = (ObjectCreationExpressionSyntax)context.Node;
+            var identifier = objectCreationExpressionSyntax.Type as QualifiedNameSyntax;
+            return identifier?.Right.ToString() != methodInformation.MethodName;
```
giggio: @andrecarlucci You are comparing by string, not by symbol. This probably can cause unnexpected mnatches. If I name something `Uri`, it will probably match on the `AbreviatedConstructorNameNotFound` method, right? This will cause incorrect diagnostics. Can we do better?

andrecarlucci:
Ok, the changes are ready on my local repo.
I'm waiting for this PR: https://github.com/code-cracker/code-cracker/pull/199 because I mistakenly created a branch from that branch, it will be a lot of work to rollback everything and submitting a new PR now would show all my changes back from the 199 again.

src/CSharp/CodeCracker/Usage/MethodAnalyzers/MethodChecker.cs
andrecarlucci: Yes, it could, but it will failed once the semantic model is loaded (MethodFullNameNotFound).

The idea is to fail fast without loading the semantic model. I don't think there's a way to do that only with the syntax tree, because you could name your namespace System.Net as well.

Do you know a way? I'll investigate this further.

andrecarlucci: It works just fine :)
https://github.com/code-cracker/code-cracker/pull/200

giggio: Ok, it makes sense to fail fast. Unless some writes a class named `System.Uri`, which would be pretty dumb.

PR #195, opened 2015-01-04
Simplify NameOfCodeFixProvider

felipebz:
CC0021 is changing:

```csharp
public class TypeName
{
    void Foo(string a, string b)
    {
        string foo = "b";
    }
}
```

To:

```csharp
public class TypeName
{
    void Foo(string a, string b)
    {
        string foo = nameof(a);
    }
}
```

The NameOfCodeFixProvider was getting always the first parameter and using its name to create the nameof expression. But we don't need to look at the parameters, because we know the parameter name is equals to the literal to be replaced.


viniciushana:
Looks great, the code is simpler too!

Issue #179, opened 2014-12-29
Interfaces should start with an "I"


giggio:
This is a convention in .NET, so:

````csharp
interface Foo { }
````

Becomes:

````csharp
interface IFoo { }
````

When renaming the interface, all references have to be renamed as well and no code should break.

Category: `Style`
Diagnostic Id: `CC0062`
Severity: `Info`

PR #196, opened 2015-01-04
InterfaceName

carloscds:
This close issue #179


giggio:
I am looking into it.

src/CSharp/CodeCracker/Usage/InterfaceNameCodeFixProvider.cs
```diff
@@ -0,0 +1,54 @@
...
+        private async Task<Document> ChangeInterfaceNameAsync(Document document, InterfaceDeclarationSyntax interfaceStatement, CancellationToken cancellationToken, FixType fixType)
+        {
+            var semanticModel = await document.GetSemanticModelAsync(cancellationToken);
+
+            var name = "I"+interfaceStatement.Identifier.ToString();
+
+            var newInterface = SyntaxFactory.InterfaceDeclaration(name);
+
+            var root = await document.GetSyntaxRootAsync();
+            var newRoot = root.ReplaceNode(interfaceStatement, newInterface);
```
giggio: @carloscds don't replace the name like that, use
````csharp
var newSolution = await Renamer.RenameSymbolAsync(document.Project.Solution, typeSymbol, newName, optionSet, cancellationToken).ConfigureAwait(false);
````
If you replace like that it will break everyone who references the interface.

File: test/CSharp/CodeCracker.Test/Style/InterfaceNameTests.cs [hunk 2]
```diff
@@ -0,0 +1,68 @@
...
+        {
+            public void Test();
+        }
+    }";
+            const string fixtest = @"
+    namespace ConsoleApplication1
+    {
+        public interface IFoo
+        {
+            public void Test();
```
giggio: This is not valid interface code, the `public` is not allowed. Same for all above.

giggio:
Can you add a test like this:

This:

````csharp
public interface MyInterface
{
    void Foo();
}
public class MyClass : MyInterface
{
    public void Foo() { }
}
````

Becomes:


````csharp
public inteface IMyInterface
{
    void Foo();
}
public class MyClass : IMyInterface
{
    public void Foo() { }
}
````

I think this will break at this point, fix that and I think we are good. Check out the rename action I commented on the code.

carloscds:
@giggio This branch is wrong. I think that git doesn´t sync my code. I will review.

giggio:
@carloscds the renaming problem persists. If you rename the interface like that you are breaking code. Change it using something like that:

````csharp
var newSolution = await Renamer.RenameSymbolAsync(document.Project.Solution, typeSymbol, newName, optionSet, cancellationToken).ConfigureAwait(false);
````

The test I provided above should help you.

carloscds:
@giggio Now implement RenameAsync(). It´s very cool now!!!

giggio:
@carloscds good job!

PR #197, opened 2015-01-05
Fix NullReferenceException in ForInArrayAnalyzer (closes #193)

### LINKED ISSUE #193: Null Reference BUG on ForInArrayAnalyzer (CC0006)
See line 313 of the build log:
https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.241#L313
To repro, just open [corefx](https://github.com/dotnet/corefx) and build.

Here is the full log:

> 1083   info AnalyzerDriver: The Compiler Analyzer 'CodeCracker.Style.ForInArrayAnalyzer' threw an exception with message 'Object reference not set to an instance of an object.'. 

**Note**:
Exact cause still to be determined, as soon as anyone investigates we can update title and description to be clearer.

felipebz:
Related to #193. When the for condition is not a BinaryExpressionSyntax, tthe code was basically doing

    if (!null ?? false) return;

And this caused a NullReferenceException in the line 44.

I just added the parentheses, it fixes the bug. But I don't think it's clear enough, I prefer to change to:

    if (condition == null || !condition.IsKind(SyntaxKind.LessThanExpression)) return;

What do you think?

### DIFF
--- a/src/CSharp/CodeCracker/Style/ForInArrayAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Style/ForInArrayAnalyzer.cs
@@ -40,7 +40,7 @@ namespace CodeCracker.Style
             var forBlock = forStatement.Statement as BlockSyntax;
             if (forBlock == null) return;
             var condition = forStatement.Condition as BinaryExpressionSyntax;
-            if (!condition?.IsKind(SyntaxKind.LessThanExpression) ?? false) return;
+            if (condition == null || !condition.IsKind(SyntaxKind.LessThanExpression)) return;
             var arrayAccessor = condition.Right as MemberAccessExpressionSyntax;
             if (arrayAccessor == null) return;
             if (!arrayAccessor.IsKind(SyntaxKind.SimpleMemberAccessExpression)) return;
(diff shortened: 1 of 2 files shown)


giggio:
I agree it is clearer. Fell free to update and we will merge it in. It is passing the build:
https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.250

giggio:
@felipebz I just noticed that it dot fix #193. You probably found some other bug. See the build:
https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.250
You will see that the error on corefx is still there (end of the file, in red).

felipebz:
I changed the if condition.

@giggio you are right. If you look at the corefx.log in the 1.0.0.241 build, you will see the error happens in System\Collections\Immutable\ImmutableArray`1.cs. Now, it is happening in System\Collections\Immutable\ImmutableArray`1+Builder.cs. It is another bug.

I will take a look at this.

felipebz:
Second bug fixed too.

giggio:
Build is still passing.
https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.254

Issue #188, opened 2015-01-04
Validate IPAddress.Parse from System.Net


andrecarlucci:
This:
`
 System.Net.IPAddress.Parse("wrong ip");
`

Will be marked as an Error. 
Category is Usage.
Diagnostic id is CC0064

andrecarlucci:
I'll do it :)

giggio:
Great!

ravpacheco:
Hi guys, 

would be great if we could supress this error.

Now if I make

    var correctIP = "0.0.0.0";
    System.Net.IPAddress.Parse(correctIP);

I receive a error.
What you think ?

giggio:
@ravpacheco yes, it is a bug. I created issue #571 to fix it.

giggio:
@ravpacheco In the meantime, you can disable it by adding:

````csharp
var correctIP = "0.0.0.0";
#pragma warning disable CC0064
System.Net.IPAddress.Parse(correctIP);
#pragma warning restore CC0064
````

ravpacheco:
@giggio great. Thanks!
Congratulations this project is awesome! \o/

PR #199, opened 2015-01-06
fixes - #188 - Validate IPAddress.Parse from System.Net

andrecarlucci:
MethodChecker <3


viniciushana:
Looks great and build passes, thank you!
However, I'm missing two things and I'd like to know your thoughts on them:

- IPv6 is also supported in `IPAddress.Parse`, it'd be nice to have some tests on this
- `IPAddress.Parse` understands strings partly formatted, e.g. translating `20.2` to `20.0.0.2` and `65536` to `0.0.255.255`

andrecarlucci:
I thought about adding those tests at the time but I didn't because that would mean testing the IPAddress.Parse (Microsoft's job :) and not our Analyzer. The Analyzer calls the original method so everything that works there will work for us too.

viniciushana:
Good point! :)

paulomorgado:
Sometimes, the most difficult thing about unit testing is figuring out what the unit is. :smile:

PR #200, opened 2015-01-06
make tests work with any culture - refs #182

andrecarlucci:
This PR make Uri tests work with any culture and also adds a unit test to validate that it won't return a diagnostic for user created classes named Uri.

### DIFF
--- a/test/CSharp/CodeCracker.Test/Usage/UriAnalyzerTests.cs
+++ b/test/CSharp/CodeCracker.Test/Usage/UriAnalyzerTests.cs
@@ -1,4 +1,5 @@
-﻿using System.Threading.Tasks;
+﻿using System;
+using System.Threading.Tasks;
 using CodeCracker.Usage;
 using Microsoft.CodeAnalysis;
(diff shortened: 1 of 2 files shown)


test/CSharp/CodeCracker.Test/Usage/UriAnalyzerTests.cs
```diff
@@ -57,17 +58,48 @@ public Person()
...
+        public Uri(string uri) { }
+
+        public void Test() {
+            var uri = new Uri(""whoCares"");
+        }
+    }
+}";
+            await VerifyCSharpHasNoDiagnosticsAsync(code);
+        }
+
```
andrecarlucci: @giggio

giggio: Great!

carloscds: Interesting!

Issue #178, opened 2014-12-29
If method returns a Task it should have the postfix "Async"


giggio:
This is a convention in .NET, so:

````csharp
Task Foo() { }
````

Becomes:

````csharp
Task FooAsync() { }
````

The `async` keyword is not necessary for the method to have the `Async` postfix.
When renaming the method, all references have to be renamed as well and no code should break.

Category: `Style`
Diagnostic Id: `CC0061`
Severity: `Info`

PR #208, opened 2015-01-09
TaskNameAsync

carloscds:
Close issue #178

### DIFF
--- a/src/CSharp/CodeCracker/Style/TaskNameAsyncAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Style/TaskNameAsyncAnalyzer.cs
@@ -36,14 +36,15 @@ namespace CodeCracker.Style
         private static void AnalyzeMethod(SyntaxNodeAnalysisContext context)
         {
             var invocationExpression = (MethodDeclarationSyntax)context.Node;
+            if (invocationExpression.Identifier.ToString().EndsWith("Async")) return;
 
-            var nameSyntax = invocationExpression.ReturnType?.ToString();
-            if (nameSyntax == null || !nameSyntax.Contains("Task")) return;
-            if(invocationExpression.Identifier.ToString().EndsWith("Async")) return;
-
-            var errorMessage = invocationExpression.Identifier.ToString()+"Async";
-            var diag = Diagnostic.Create(Rule, invocationExpression.GetLocation(),errorMessage);
-            context.ReportDiagnostic(diag);
+            var returnTask = invocationExpression.ReturnType?.ToString().Contains("Task") ?? false;
+            if (invocationExpression.Modifiers.Any(SyntaxKind.AsyncKeyword) || returnTask)
+            {
+                var errorMessage = invocationExpression.Identifier.ToString() + "Async";
+                var diag = Diagnostic.Create(Rule, invocationExpression.GetLocation(), errorMessage);
+                context.ReportDiagnostic(diag);
+            }
         }
     }
 }
--- a/src/CSharp/CodeCracker/Style/TaskNameAsyncAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Style/TaskNameAsyncAnalyzer.cs
@@ -12,7 +12,7 @@ namespace CodeCracker.Style
     {
         public const string DiagnosticId = "CC0061";
         internal const string Title = "Async method can be terminating with 'Async' name.";
-        internal const string MessageFormat = "Change method name including 'Async'.";
+        internal const string MessageFormat = "Change method name to {0}";
         internal const string Category = SupportedCategories.Style;
         const string Description = "Async method can be terminating with 'Async' name.";
 
@@ -30,24 +30,19 @@ namespace CodeCracker.Style
 
         public override void Initialize(AnalysisContext context)
         {
-            context.RegisterSyntaxNodeAction(AnalyzeClass, SyntaxKind.MethodDeclaration);
+            context.RegisterSyntaxNodeAction(AnalyzeMethod, SyntaxKind.MethodDeclaration);
         }
 
-        public Task<bool> xpto()
-        {
-            return null;
-        }
-
-        private static void AnalyzeClass(SyntaxNodeAnalysisContext context)
+        private static void AnalyzeMethod(SyntaxNodeAnalysisContext context)
         {
             var invocationExpression = (MethodDeclarationSyntax)context.Node;
 
             var nameSyntax = invocationExpression.ReturnType?.ToString();
             if (nameSyntax == null || !nameSyntax.Contains("Task")) return;
-            if(invocationExpression.Identifier.ToString().ToUpper().EndsWith("ASYNC")) return;
+            if(invocationExpression.Identifier.ToString().EndsWith("Async")) return;
 
-            var error = string.Format(MessageFormat, MessageFormat);
-            var diag = Diagnostic.Create(Rule, invocationExpression.GetLocation(), error);
+            var errorMessage = invocationExpression.Identifier.ToString()+"Async";
+            var diag = Diagnostic.Create(Rule, invocationExpression.GetLocation(),errorMessage);
             context.ReportDiagnostic(diag);
         }
     }
--- a/src/CSharp/CodeCracker/Style/TaskNameAsyncAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Style/TaskNameAsyncAnalyzer.cs
@@ -1,0 +1,54 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+
+namespace CodeCracker.Style
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class TaskNameAsyncAnalyzer : DiagnosticAnalyzer
+    {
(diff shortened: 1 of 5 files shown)


giggio:
Build is passing: https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.269

src/CSharp/CodeCracker/Style/TaskNameAsyncAnalyzer.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterSyntaxNodeAction(AnalyzeClass, SyntaxKind.MethodDeclaration);
+        }
+
+        public Task<bool> xpto()
+        {
+            return null;
+        }
```
giggio: @carloscds This code should be removed. :)

src/CSharp/CodeCracker/Style/TaskNameAsyncAnalyzer.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        private static void AnalyzeClass(SyntaxNodeAnalysisContext context)
+        {
+            var invocationExpression = (MethodDeclarationSyntax)context.Node;
+
+            var nameSyntax = invocationExpression.ReturnType?.ToString();
+            if (nameSyntax == null || !nameSyntax.Contains("Task")) return;
```
giggio: Don't check for the `Task` name, check for the type using semantic analysis. Remember that we don't have to return a `Task`, I can return anything that inherits from it, and that would not have the `Task` name. Actually, we could even work the awaitable pattern, but that would be too much at this point, looking into anything that inherits from `Task` or `Task` itself would be enough.

src/CSharp/CodeCracker/Style/TaskNameAsyncCodeFixProvider.cs
```diff
@@ -0,0 +1,50 @@
...
+            return WellKnownFixAllProviders.BatchFixer;
+        }
+
+        public sealed override async Task ComputeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var diagnostic = context.Diagnostics.First();
+            var diagnosticSpan = diagnostic.Location.SourceSpan;
+            var declaration = root.FindToken(diagnosticSpan.Start).Parent.AncestorsAndSelf().OfType<MethodDeclarationSyntax>().First();
+            context.RegisterFix(CodeAction.Create("Change method name including 'Async'.", c => ChangeMethodNameAsync(context.Document, declaration, c)), diagnostic);
```
giggio: I would prefer if you used the method name on the code fix description, it is very nice, like `Change method name to 'FooAsync'`. What do you think?

src/CSharp/CodeCracker/Style/TaskNameAsyncAnalyzer.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            context.RegisterSyntaxNodeAction(AnalyzeClass, SyntaxKind.MethodDeclaration);
+        }
+
+        public Task<bool> xpto()
+        {
+            return null;
+        }
+
+        private static void AnalyzeClass(SyntaxNodeAnalysisContext context)
```
giggio: SHouldn't this method be named `AnalyzeMethod`?

src/CSharp/CodeCracker/Style/TaskNameAsyncAnalyzer.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            var invocationExpression = (MethodDeclarationSyntax)context.Node;
+
+            var nameSyntax = invocationExpression.ReturnType?.ToString();
+            if (nameSyntax == null || !nameSyntax.Contains("Task")) return;
+            if(invocationExpression.Identifier.ToString().ToUpper().EndsWith("ASYNC")) return;
```
giggio: I would check for an exact `"Async"` match, case sensitive.

src/CSharp/CodeCracker/Style/TaskNameAsyncAnalyzer.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Threading.Tasks;
+
+namespace CodeCracker.Style
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class TaskNameAsyncAnalyzer : DiagnosticAnalyzer
+    {
+        public const string DiagnosticId = "CC0061";
+        internal const string Title = "Async method can be terminating with 'Async' name.";
+        internal const string MessageFormat = "Change method name including 'Async'.";
```
giggio: Why not `Change method name to FooAsync`, using a `MessageFormat` like `Change method name to {0}`?

src/CSharp/CodeCracker/Style/TaskNameAsyncCodeFixProvider.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        private async Task<Solution> ChangeMethodNameAsync(Document document, MethodDeclarationSyntax methodStatement, CancellationToken cancellationToken)
+        {
+            var semanticModel = await document.GetSemanticModelAsync(cancellationToken);
+            var newName = methodStatement.Identifier.ToString()+"Async";
+        
+            var solution = document.Project.Solution;
+            if (solution == null) return null;
+            var symbol = semanticModel.GetDeclaredSymbol(methodStatement, cancellationToken);
+            if (symbol == null) return null;
```
giggio: No need for the null checks on the code fix, it has already been looked at by the analyzer. You can remove all of them.

giggio:
I saw that it works also with interface and when you update an interface method it updates the implementation as well, which is nice. How about adding a test covering it?

src/CSharp/CodeCracker/Style/TaskNameAsyncAnalyzer.cs
carloscds: I want that I check modifier and not returnType ? Why ? Give a sample that show your ideia.

src/CSharp/CodeCracker/Style/TaskNameAsyncCodeFixProvider.cs
carloscds: Code preview already show this.

carloscds:
@giggio Done.

giggio:
Cool. I will check it out. Build is passing: https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.276

src/CSharp/CodeCracker/Style/TaskNameAsyncAnalyzer.cs
giggio: Not the modifier, @carloscds, because an async method might not have the `async` modifier.
But consider this:

````csharp
class MyPromise : Task {}
class Foo
{
    async MyPromise DoSomething()
    {
    }
}
````

This method should actually be called `DoSomethingAsync`, but currently it won't be detected, as the return type is called `MyPromise`, and does not have `Task` in its name.
You have to check if the return type is a `Task`, or if it inherits from `Task`.

Check this method I added, `AllBaseTypesAndSelf`:
https://github.com/giggio/code-cracker/blob/unused-parameters/src/CSharp/CodeCracker/Extensions/AnalyzerExtensions.cs#L152

See also how it is used. Is it not in yet, as we are still discussing that PR, but that is the idea.

carloscds: Ok. It´s make sense.

carloscds:
@giggio Done.

src/CSharp/CodeCracker/Style/TaskNameAsyncAnalyzer.cs
giggio: @carloscds You still need to work this case for the PR to get in.

giggio:
@carloscds This should generate a diagnostic and be fixable:

````csharp
class MyPromise : Task {}
class Foo
{
    async MyPromise DoSomething()
    {
    }
}
````

Becomes:

````csharp
class MyPromise : Task {}
class Foo
{
    async MyPromise DoSomethingAsync()
    {
    }
}
````

Checkout the `AllBaseTypesAndSelf` method from Analyzer extensions. You just need to pass the return type symbol. It can help:

https://github.com/code-cracker/code-cracker/blob/master/src/CSharp/CodeCracker/Extensions/AnalyzerExtensions.cs#L152

carloscds:
@giggio Your sample code already work with this analyzer since last week. Please test again.

giggio:
@carloscds Ok, 2 things:

The test case I had posted before is incorrect. You cannot inherit from `Task` and use that on an async method. Section 10.15 of the C# language spec says that:

> The return-type of an async method must be either void or a task type. The task types are System.Threading.Tasks.Task and types constructed from System.Threading.Tasks.Task<T>. For the sake of brevity, in this chapter these types are referenced as Task and Task<T>, respectively. An async method returning a task type is said to be task-returning.

So, forget that. Sorry about that.

The second thing is that, because of this line on `TaskNameAsyncAnalyzer`:

````csharp
var returnTask = invocationExpression.ReturnType?.ToString().Contains("Task") ?? false;
````

A diagnostic is being raised for this method `Bar`, and it shouldn't:

````csharp
class Foo
{
  ATask Bar()
  {
  }
}
class ATask { }
````

Can you fix that? You should be checking if the return type is `Task` or `Task<T>`, like so:

````csharp
semanticModel.Compilation.GetTypeByMetadataName("System.Threading.Tasks.Task")
````

And then checking for `.Equals`.

Issue #209, opened 2015-01-11
BUG on UriAnalyzer when analyzing an expression


giggio:
This code:

````csharp
            var mainConstrutor = new MethodInformation(
                "Uri",
                "System.Uri.Uri(string)",
                args => { new Uri(args[0].ToString()); }
            );
````

Is raising an error diagnostic on `CC0063`. It shows a `NullReferenceException` on the `new Uri(args[0].ToString())` statement. I am guessing the `UriAnalyzer` is trying to call `new Uri` and can't find `args`.
I have disabled the analyzer on our dogfooding.

giggio:
@andrecarlucci this one is your baby. I would like you to fix it, because you know the internals better than anyone. What do you think?

PR #210, opened 2015-01-11
IPAdressAnalyzer diagnostic id was wrong. CC0063 instead CC0064 - refs #209

andrecarlucci:
@giggio, the IPAdressAnalyzer was using Uri's diagnostic Id. Do you think that could be the source of the error?


giggio:
I don't know, @andrecarlucci. I don't think so. Open CC on a VS Roslyn instance (run the project) and see if the still points that error.

giggio:
@andrecarlucci I just tested it. It still fails. Just add this test:

````csharp
new Uri(args[0].ToString());
````

PR #212, opened 2015-01-11
fixes #209 - BUG on UriAnalyzer

### LINKED ISSUE #209: BUG on UriAnalyzer when analyzing an expression

andrecarlucci:
I changed the analyzer to ignore when the arg is null.
I could change MethodChecker to do that always, but since some methods do accept null (arg!), I let this responsibility to the caller.

### DIFF
--- a/src/CSharp/CodeCracker/Usage/UriAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Usage/UriAnalyzer.cs
@@ -43,12 +43,26 @@ namespace CodeCracker.Usage
             var mainConstrutor = new MethodInformation(
                 "Uri",
                 "System.Uri.Uri(string)",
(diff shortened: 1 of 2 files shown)


giggio:
Build is passing.
https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.279

Issue #214, opened 2015-01-11
CodeCracker Base Classes?


AdamSpeight2008:
The base class for all CodeCracker Analyzers

```VB
MustInherit Class CodeCracker_Analyzer
  Inherits CodeAnalyzer

  Public ReadOnly Property DiagnosticID As String = ""
  Public ReadOnly Property Title As String = ""
  Public ReadOnly Property Category As String = ""
  Public ReadOnly Property Severity As DiagnosticSeverity = DiagnosticSeverity.Warning
  Public ReadOnly Property MsgFormat As String = ""
  Public ReadOnly Propery IsEnable As Boolean = True

  Friend _Diagnostics As ImmutabaleArray(Of DiagnosticDescriptor)

  Friend Sub New( ID As String, Title As String, MsgFormat As String, Category As String,
                    Optional IsEnable As Boolean = True,
                    Optional Severity As DiagnosticSeverity = DiagnosticSeverity.Warning)

    MyBase.New()
    Me._DiagnosticID = ID
    Me._Title =  "ForInArrayAnalyzer") 
    Me._Category = Category 
    Me._MsgFormat = MsgFormat
    Me._Severity = Severity
    Me._IsEnabled = IsEnabled
  End Sub
  
  Public Overrides ReadOnly Property SupportedDiagnostics() As ImmutableArray(Of DiagnositcDescriptor)
    Get
      Return _Diagnostics
    End Get
  End Property

  Friend Function GetDescriptor() As DiagnositcDescriptor
    Return New DiagnosticDescriptor( DiagnosticID, Title, MsgFormat, Category, Severity, IsEnabledByDefault, HelpLink.ForDiagnostic( DiagnosticID ) )
  End Function

 Public Override Sub Initialize(context As AnalysisContext )
   context.RegisterSyntaxNodeAction(Analyzer, SyntaxKind.CatchClause);
 End Sub

End Class
```
Now a base class for CSharp Analyzers

```VB
<DiagnosticAnalyzer(LanguageNames.CSharp)>
Public MustInheric CodeCracker_CSharp
  Friend Sub New( ID As String, Title As String, MsgFormat As String, Category As String,
                    Optional IsEnable As Boolean = True,
                    Optional Severity As DiagnosticSeverity = DiagnosticSeverity.Warning)

    MyBase.New()
    Me._DiagnosticID = ID
    Me._Title =  "ForInArrayAnalyzer") 
    Ne._Category = Category 
  End Sub
End Sub
```
[...]

PR #221, opened 2015-01-17
Base classes

giggio:
Related to #214

This is only a start so we can discuss. Please **don't accept the PR** until we are finished discussing.
@ElemarJR @viniciushana @carloscds I need your feedback on this work on base classes.

### DIFF
--- a/src/CSharp/CodeCracker/Analyzer.cs
+++ b/src/CSharp/CodeCracker/Analyzer.cs
@@ -1,0 +1,40 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Text;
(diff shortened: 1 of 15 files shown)


carloscds:
I think that this approach will be left more complex to begninners create analyzers.

carloscds:
I didn´t agree with this approache, it´s not necessary create bases classes to create analyzers. Keep simplicity!

giggio:
@code-cracker/owners I have discussed the issue with @carloscds and we agreed that the enum to hold the diagnostic ids is a good idea. I will go ahead and implement that.
We still need to discuss the rest. Let's schedule a meting and decide what we will do.

carloscds:
@code-cracker/owners Great! Let me know if you need help.

codecrackerbuild:
Teste

giggio:
We can't find any more things from this PR to bring to CC.
We see too much complexity on creating a class hierarchy at the moment. We prefer to keep it simple.

PR #223, opened 2015-01-18
TaskNameAsync

carloscds:
I have created this new PR because new version of VS 2015.


giggio:
@carloscds `TaskNameAsyncAnalyzer` is throwing a null reference exception on line 43, `returnType` is null, if you use a method with an return type not yet defined:

````csharp
Class Foo
{
  ClassThatDoesNotExist Foo()
  {
  }
}
````

Also, anything that matches the namespace `System.Threading.Tasks` is raising the dianostic, you have to check specifically for `Task` and `Task<T>`. This will incorrectly create a diagnostic:

````csharp
class Foo
{
	TaskFactory DoSomething()
	{
	}
}
````

You should also add a test for that.

carloscds:
@giggio First case corrected. Second case is more complex than this, because has too: Task<bool>, Task<string>, Task<wathever...>

giggio:
@carloscds you could do:

````csharp
returnType == "System.Threading.Tasks.Task" || returnType == "System.Threading.Tasks.Task<T>"
````

I am not sure about Task<T> you will have to check. Add a test and see what the text is.

carloscds:
@giggio Done!

giggio:
@carloscds we can't use `returnType?.ConstructedFrom.ToString().StartsWith("System.Threading.Tasks.Task")`, because it will match `TaskFactory`, `TaskScheduler`, etc, and that will be incorrect.

Also, `invocationExpression` is not a good name for a `MethodDeclarationSyntax`, agree?

carloscds:
@giggio Done!

giggio:
Build is passing: https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.314

giggio:
@carloscds Merged! \o/

Issue #225, opened 2015-01-22
BUG on CC0026: When there is a dynamic don't raise a diagnostic


giggio:
This code:

````csharp
Microsoft.CodeAnalysis.CSharp.CSharpExtensions.TryGetSpeculativeSemanticModel(semanticModel, originalExpression.SpanStart, (dynamic)GetSemanticRootForSpeculation(replacedExpression), out speculativeModel)
````

Will raise CC0026 incorrectly.

If you call it like an extension method, like this:

````csharp
semanticModel.TryGetSpeculativeSemanticModel(originalExpression.SpanStart, (dynamic)GetSemanticRootForSpeculation(replacedExpression), out speculativeModel)
````

This call will fail because of the dynamic argument that will not allow binding to the extension method on runtime.

@sqdavid is working on it.

giggio:
CC0026 is now disabled by default until we fix it.

sqdavid:
Maybe this test can help

```csharp
[Fact]
public async Task WhenCallExtensionMethodWithDynamicArgumentHasNoDiagnostics()
{
    const string source = @"
using System;
using System.Collections.Generic;
using System.Linq;

namespace ConsoleApplication1
{
    class ExtensionTest
    {
        void method()
        {
            var list = new List<int>() { 5, 56, 2, 4, 63, 2 };
            dynamic dList = list;
            Console.WriteLine(Enumerable.First(dList)); 
            //Console.WriteLine(dList.First());
        }
    }
}
";
    await VerifyCSharpHasNoDiagnosticsAsync(source);
}

sqdavid:
If nobody is working on it, I have solved it.

carloscds:
@sqdavid Go ahead.

sqdavid:
I've just created the PR #231 
Please review it.

PR #231, opened 2015-01-24
Verify arguments' TypeInfo to identify dynamic arguments

sqdavid:
BUG on CC0026: When there is a dynamic don't raise a diagnostic #225


giggio:
Nice!

PR #239, opened 2015-01-29
Issues 236 & 238

thomaslevesque:


src/CSharp/CodeCracker/Usage/SimplifyRedundantBooleanComparisonsAnalyzer.cs
```diff
@@ -31,18 +31,29 @@ public override void Initialize(AnalysisContext context)
...
-            var right = context.SemanticModel.GetConstantValue(comparison.Right);
-            if (!right.HasValue) return;
+            // Only handle the case where both operands of type bool; other cases involve
+            // too much complexity to be able to deliver an accurate diagnostic confidently.
+            var leftType = context.SemanticModel.GetTypeInfo(comparison.Left).Type;
+            var rightType = context.SemanticModel.GetTypeInfo(comparison.Right).Type;
+            if (!IsBoolean(leftType) || !IsBoolean(rightType))
+                return;
 
-            if (!(right.Value is bool)) return;
```
thomaslevesque: This test is made unnecessary by previous check - we already know that it's a bool

giggio:
@thomaslevesque Your update is throwing an NullReferenceException when analyzing Corefx: https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.341
Run `\test\CSharp\AnalyzeCoreFx.ps1` on your local machine, it will leave a temp directory with corefx and the analyzers installed and you can investigate.

src/CSharp/CodeCracker/Usage/SimplifyRedundantBooleanComparisonsAnalyzer.cs
```diff
@@ -31,18 +31,29 @@ public override void Initialize(AnalysisContext context)
...
         {
             var comparison = (BinaryExpressionSyntax)context.Node;
-            var semanticModel = context.SemanticModel;
 
-            var right = context.SemanticModel.GetConstantValue(comparison.Right);
-            if (!right.HasValue) return;
+            // Only handle the case where both operands of type bool; other cases involve
+            // too much complexity to be able to deliver an accurate diagnostic confidently.
+            var leftType = context.SemanticModel.GetTypeInfo(comparison.Left).Type;
+            var rightType = context.SemanticModel.GetTypeInfo(comparison.Right).Type;
```
giggio: Maybe `leftType` or `rightType` are null and then it is throwing the nullref.

thomaslevesque: I don't know how it's possible... If the type is unknown, `TypeInfo.Type` returns a `IErrorTypeSymbol`, not null. But perhaps it's `GetTypeInfo` that returns null... I'll investigate

thomaslevesque:
I'm unable to reproduce the error on my machine... AnalyzeCoreFx.ps1 runs without errors, and the log contains no mention of SimplifyRedundantBooleanComparisonsAnalyzer. However I noticed this in corefx.log (several times):

> Target "RunCodeAnalysis" skipped, due to false condition; ('$(RunCodeAnalysisOnThisProject)'=='true') was evaluated as (''=='true').

So it looks like the analysis doesn't even run... Any idea what could cause this?

thomaslevesque:
OK, I was using the wrong MSBuild version... got the error now

thomaslevesque:
You were right, there is a case where leftType or rightType can be null: when one of the operands is the literal `null`. Haven't been able to narrow down the exact scenario where this happens, but it's easy to check that it isn't null. I fixed it.

AdamSpeight2008:
Does it correctly not report in the cases where the condition is incomplete?
`if(  == )` or `if( a == )` or `if( == b)` ?

thomaslevesque:
@AdamSpeight2008, good point. It correctly ignores that case; I just added a test to make sure.

giggio:
Ok, it is fixed now: https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.342

PR #242, opened 2015-01-30
Implemented UseConfigureAwaitFalse analyzer and code fix

### LINKED ISSUE #235: Use ConfigureAwait(false) on awaited task
Best practice dictates that `ConfigureAwait(false)` should always be used in library code, unless the async operation specifically needs to resume on the original `SynchronizationContext`. So I think it would be useful to create an analyzer for this.

- Level: not sure whether the level should be *Information* or *Warning*.
- Code fix: obviously, call `ConfigureAwait(false)` on the awaited task

**Scenario**

    await FooAsync();

becomes 

    await FooAsync().ConfigureAwait(false);

Another thing to consider: should the analyzer warn only on the first awaited task in the method, or all of them? Unless I'm mistaken, if `ConfigureAwait(false)` is used on the first `await`, it's not necessary to use it on the others, since we have already "lost" the `SynchronizationContext`, so there isn't a SC to capture anyway.

Diagnostic Id: `CC0070`
Severity: `Hidden` (Refactoring)
Category: `Reliability`

thomaslevesque:
Fix #235

### DIFF
--- a/src/CSharp/CodeCracker/CodeCracker.csproj
+++ b/src/CSharp/CodeCracker/CodeCracker.csproj
@@ -43,6 +43,8 @@
     <Compile Include="Refactoring\AllowMembersOrderingCodeFixProvider.Base.cs" />
     <Compile Include="Refactoring\ParameterRefactoryAnalyzer.cs" />
     <Compile Include="Refactoring\ParameterRefactoryCodeFixProvider.cs" />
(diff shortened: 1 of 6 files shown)


giggio:
Build is passing: https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.344.

giggio:
Overall it looks good, it is pretty late now in Brazil, I will finish checking it tomorrow.

PR #259, opened 2015-02-06
IntroduceFieldConstructor

carloscds:


giggio:
@carloscds we are not using `_` on CodeCracker for anything, if we accept it like that, we won't be able to use this analyzer on CodeCracker itself. :(
The problem will present itself anyway. If you get this scenario:

````csharp
class Foo
{
  private readonly int _bar;
  public Foo(string bar)
  {
  }
}
````

You still would have to deal with the name collision.
My suggestion is: remove the `_`, use the numbers we had agreed on before.

carloscds:
@giggio Although I guess that underscore is not a problem. I changed fixer to original proposal.

giggio:
@carloscds I think you had a problem with your commit. This is your latest commit: eda6daf6c61dcd791d56021a9c0369fa580ca26f
Code fix provider is broken.

jwooley:
I would recommend NOT including the underscore analyzer for VB versions because underscore is frequently used for privates in VB since VB doesn't allow disambiguation by casing. In fact, the VB compiler actually uses the underscore for the private backing field on auto-properties. See https://msdn.microsoft.com/en-us/library/dd293589.aspx.

carloscds:
@jwooley This was removed and not I'm using a number for field.

Issue #241, opened 2015-01-30
Introduce field from constructor


giggio:
This is a refactoring.

The idea is that on a constructor parameter you CTRL DOT and get a "Introduce field" dialog, that will create a field and assign the value from the parameter to the field.

So this:

````csharp
class Foo
{
  public Foo(string bar)
  {
  }
}
````

Becomes:

````csharp
class Foo
{
  private readonly string bar;
  public Foo(string bar)
  {
    this.bar = bar;
  }
}
````

If the field is already assigned then no diagnostic is offered.
If the field is already present, it is just assigned to.
If the type does not match (e.g. parameter is `string`, existing field is `int`, a new field with `1` postfixed is added, like that:

````csharp
class Foo
{
  private int bar;
  public Foo(string bar)
  {
  }
}
````

Becomes:

````csharp
class Foo
{
  private int bar;
  private readonly string bar1;
  public Foo(string bar)
  {
    this.bar1 = bar;
  }
}
````

Diagnostic Id: `CC0071`
Severity: `Hidden` (refactoring)
Category: `Refactoring`

carloscds:
@giggio It's mine!

giggio:
I knew this one was not going to stay up for too long. It is too useful. :)

carloscds:
@giggio It's done. But I have a little change:

So this:

````csharp
class Foo
{
  public Foo(string bar)
  {
  }
}
````

Becomes:

````csharp
class Foo
{
  private readonly string _bar;
  public Foo(string bar)
  {
    this._bar = bar;
  }
}
````
This avoid change name of parameter if there is a field with same name.

giggio:
@carloscds I don't agree with the underline. I will comment on the PR.

carloscds:
@giggio I have submit a fix for this and a comment on the PR.

PR #261, opened 2015-02-07
Introduce field from constructor #241

carloscds:


test/CSharp/CodeCracker.Test/Refactoring/IntroduceFieldFromConstructorTest.cs
```diff
@@ -0,0 +1,149 @@
...
+            private int par;
+
+            public TypeName(int par)
+            {
+               this.par1 = par;
+            }
+        }
+    }";
+            await VerifyCSharpFixAsync(test, expected);
+        }
```
giggio: @carloscds I think that in this case it should assign `this.par = par`, as the name and type match.
It should only create `par1` if its type was not `int`. What do you think?

src/CSharp/CodeCracker/Refactoring/IntroduceFieldFromConstructorAnalyzer.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        public override void Initialize(AnalysisContext context) => context.RegisterSyntaxNodeAction(AnalyzeConstructor, SyntaxKind.ConstructorDeclaration);
+
+        private static void AnalyzeConstructor(SyntaxNodeAnalysisContext context)
+        {
+            var constructorMethod = (ConstructorDeclarationSyntax)context.Node;
+            var parameters = constructorMethod.ParameterList.Parameters;
+
+            if (constructorMethod.Body == null) return;
+            if (constructorMethod?.Body.Statements.Count == 0 && parameters.Count > 0)
```
giggio: Check for `constructorMethod?` should be on line 38, like this:

````csharp
if (constructorMethod?.Body == null) return;
````

And line 39 does not need the `?`.

File: src/CSharp/CodeCracker/Refactoring/IntroduceFieldFromConstructorAnalyzer.cs [hunk 3]
```diff
@@ -0,0 +1,79 @@
...
+            {
+                foreach (var par in parameters)
+                {
+                    var errorMessage = par.Identifier.Text;
+                    var diag = Diagnostic.Create(Rule, constructorMethod.GetLocation(), errorMessage);
+                    context.ReportDiagnostic(diag);
+                }
+                return;
+            }
```
giggio: Very smart to take a short route and just report on every parameter if there are no statements!

src/CSharp/CodeCracker/Refactoring/IntroduceFieldFromConstructorAnalyzer.cs
```diff
@@ -0,0 +1,79 @@
...
+                        {
+                            var right = assign.Right as IdentifierNameSyntax;
+                            if (right.Identifier.Text == par.Identifier.Text)
+                            {
+                                foundAssignment = true;
+                                break;
+                            }
+                        }
+                    }
+                }
```
giggio: This is not going to work if the assignment is inside an if statement.
Maybe you should use a data flow analysis?

src/CSharp/CodeCracker/Refactoring/IntroduceFieldFromConstructorCodeFixProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            var newConstructor = constructorStatement.WithBody(constructorStatement.Body.AddStatements(assignmentField));
+            var newClassConstructor = oldClass.ReplaceNode(constructorStatement, newConstructor);
+            var newClass = newClassConstructor.WithMembers(newClassConstructor.Members.Insert(0, newField))
+                           .WithoutAnnotations(Formatter.Annotation);
+            var newRoot = root.ReplaceNode(oldClass, newClass);
+            return document.WithSyntaxRoot(newRoot);
+        }
+    }
+}
```
giggio: I guess this will not work with Fix all. Can you test that and make sure? Check the

src/CSharp/CodeCracker/Refactoring/IntroduceFieldFromConstructorAnalyzer.cs
carloscds: @giggio I need parameter name in diagnostic because I use it in analyzer.

src/CSharp/CodeCracker/Refactoring/IntroduceFieldFromConstructorAnalyzer.cs
carloscds: @giggio Maybe dataflow can solve this. I will check if is better or not.

carloscds:
@giggio Code improved!

giggio:
@carloscds I have added another commit for some cases, see the tests I added.
Also, fix all is not working and there is a name clash we need to fix. See #266 and #267. Do you want to fix those?

PR #264, opened 2015-02-09
Translate ConfigureAwait fix for VB

jwooley:


giggio:
Build is passing: https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.379

src/VisualBasic/CodeCracker/CodeCracker.vbproj
```diff
@@ -126,7 +128,6 @@
     <Folder Include="Mobility\" />
     <Folder Include="Naming\" />
     <Folder Include="Portability\" />
-    <Folder Include="Reliability\" />
```
giggio: Don't remove the .gitkeep files and the folders until we have an analyzer in them. I will add them back, but be careful in the next ones.

giggio: I just noticed that that is exactly what you did. My bad. Sorry.

PR #277, opened 2015-02-15
RemoveMethodAsync

### LINKED ISSUE #245: If method does not return a Task it shouldn't end with "Async"
This is the opposite of #178:

````csharp
void FooAsync() { }
````

Becomes:

````csharp
void Foo() { }
````

The `async` keyword is not necessary for the method to have the `Async` postfix. All it needs to be async is to return a `Task`.
When renaming the method, all references have to be renamed as well and no code should break.
Should work on interfaces as well.

Category: `Style`
Diagnostic Id: `CC0072`
Severity: `Info`

carloscds:
Close #245


giggio:
Build is passing: https://ci.appveyor.com/project/code-cracker/code-cracker/build/1.0.0.398

PR #285, opened 2015-02-21
Fix RemovePrivateMethodNeverUsed

felipebz:
Fix RemovePrivateMethodNeverUsed to correctly identify all references to the private method.

This code causes CC0068:

````csharp
using System;

public class Foo
{
    public void PublicFoo()
    {
        Action method = PrivateFoo;
    }

    private void PrivateFoo() { }
}
````

The PrivateFoo method can't be removed in this context.

(This also affects PR #284)


giggio:
Fixed on VB as well.

Issue #266, opened 2015-02-09
BUG on CC0071 (introduce field), name clash


giggio:
When there is some other identifier that is not a field CC0071 will introduce a name clash.

This will generate a field named `a` and it shouldn't:

````csharp
class Foo
{
  public event Action a;
  Foo(int a)
  {
  }
}
````

giggio:
closed on #286

PR #286, opened 2015-02-22
BUG on CC0071 (introduce field), name clash #266

carloscds:


test/CSharp/CodeCracker.Test/Refactoring/IntroduceFieldFromConstructorTest.cs
```diff
@@ -128,11 +128,12 @@ namespace ConsoleApplication1
     {
         class TypeName
         {
+            private readonly int par1;
             private readonly int par;
 
             public TypeName(int par)
             {
-               this.par = par;
+               this.par1 = par;
```
giggio: @carloscds This is not supposed to happen. The argument `par` and the field `par` have the same and can be merged together. This test `FieldAlreadyExistsAndMatchesType` verified that.
The bug only happens when you have an event or some other members, like that:

````csharp
class Foo
{
  public event Action a;
  Foo(int a)
  {
  }
}
````

carloscds:
@giggio It's not a proble, look at full code below:

````csharp
class TypeName
        {
            private readonly int par1;
            private readonly int par;
            public TypeName(int par)
            {
               this.par1 = par;
            }
        }
````
Now I will ever create a new var, because I'm checking events, methods, properties and fields. And I'm not checking specifically for types. I removed this test now and update branch.

giggio:
@carloscds This is not how it should go. If the type and the name match, it should simply use the variable, not create a new one.

giggio:
@carloscds Please update to RC1 using current HEAD.

carloscds:
@giggio I will

carloscds:
@giggio Bug Fixed.

giggio:
@carloscds also added some tests and support for inner classes, structs, interfaces, enums, and delegates.

Issue #177, opened 2014-12-29
Detect read-only not private fields and fix adding the "readonly" modifier


giggio:
Related to #86 which is already solved and is specific to **private** fields. This one is specific to non private fields.

We have to verify that the field is not set according to the accessibility of field before offering the diagnostic: if it public in the whole solution, if it is protected on the whole solution only on inherited classes, internal only on the project and protected internal only on the inherited classes on the project.
In no case should the diagnostic be reported if it would break the code when the code fix (making the field read only) is applied. 

This:

````csharp
class Foo
{
    public int a = 1;
}
````

Becomes:

````csharp
class Foo
{
    public readonly int a = 1;
}
````

See more test cases on test class `CodeCracker.Tests.Usage.ReadonlyFieldTests`.

This is only an analyzer, the code fix is already written for private fields and should work. This is also the reason we are using the same diagnostic id.

This is an `Info`.
Category is `Usage`
Diagnostic id: `CC0074`

@dmgandini is working on it.

dmgandini:
I believe the second is most common scenario as well. I want to work on it.

giggio:
@dmgandini Ok, it is yours.
Only to let you know, this is a hard one. You **cannot** `.Wait` or `.Result` on an analyzer, ok?
The Roslyn team has given us some clues on how to work on this one. You have to register an analysis on compilation start, collect all the references, then on compilation end you verify if the field was ever written to. So this will only run on full compilations, not when the user is typing. This seems to be a limitation on this specific analysis that we want to do.
I will mark it as working.

dmgandini:
I'm aware that is hard one. I was thinking to use SymbolFinder after load the current workspace, I was trying to figure out how to do it without use the MSBuildWorkspace once the project is a portable one. 
After your post I will try to do as you said. Thanks for the tips. If I get stuck I will let you guys know.

The diagnostic id, CC0052, is the same of ReadonlyFieldAnalyzer. Is it Ok?

giggio:
Don't use the Workspace, as it is not portable, and I don't think it will ever be portable.
Yes, same diagnostic id, it is pretty much the same diagnostic, just a different scope. We could add a new one, but I am not sure if it makes sense. What do you think?

giggio:
@dmgandini How is this coming? Need help?

dmgandini:
Hi @giggio!
I had some personal issues that drained my time, although I'm still working on it.
[...]

PR #292, opened 2015-03-01
Detect read-only not private fields and fix adding the "readonly" modifier #177

dmgandini:


carloscds:
@dmgandini I think that it´s a conflict between two analyzers: ReadonlyFieldAnalyzer and NoPrivateReadonlyFieldAnalyzer. For example, this code:

''''csharp
public class Foo
    {
        private int i = 0;
    }
''''

It´s analyzed by both analyzers, but only ReadonlyFieldAnalyzer create diagnostic and trigger fix, although there is a test that says that this is not an error for analyzer (PrivateFieldWithAssignmentOnDeclarationCreatesNoDiagnostic). Test is using only NoPrivateReadonlyFieldAnalyzer, but executing project and creating a code, both analyzers are running.

Can you check this ?

dmgandini:
@carloscds It is the expected behavior as far as I understood the issue. 
The ReadonlyFieldAnalyzer performs tests against private fields, because its private it needs to search probably in a single type. 

The NoPrivateReadonlyFieldAnalyzer will chase the other modifiers, because it depends to search through other types. 

I can argue that NoPrivateReadonlyFieldAnalyzer could do the both roles if CanBecameReadOnlyField method

giggio:
@dmgandini The diagnostic id is wrong, number 74 was assigned to another issue. @carloscds be careful to check the [wiki](https://github.com/code-cracker/code-cracker/wiki/DiagnosticIds). I have updated the wiki so fix this: https://github.com/code-cracker/code-cracker/wiki/DiagnosticIds/_compare/47cbc9af8941a664f3ffb85c9d599b4976b37ec1...d1353edefb9a428e43955defd731309bf04346cf

giggio:
I am not sure if the best implementation here is to the analyzis manually. Probably it would be better to use `SemanticModel.AnalyzeDataFlow`. I will open an issue to review this.

giggio:
@dmgandini `AnalyzeDataFlow` is not working for out/ref parameters. So I will leave it as is by now.

Another issue: I have disabled the analyzer for another reason: as it runs very late on the compilation stage (only when the compilation ends), it might happen that a field no longer can be made readonly, and the diagnostic remains. This is a problem, as the user might run the code fix, and that might actually break the code.
I am going to open a bug to work on the code fix for that specific case. The diagnostic actually has to run again before the code fix is done, and maybe not offer a fix at all if the user has changed something and the compilation end event did not run yet.
BTW, any idea on how to solve this problem?

Also, I have changed the registration of the compilation end to happen on the compilation start. See this line:
https://github.com/code-cracker/code-cracker/blob/d8a4d88e0016bc3cd1aa61f8d544a9412004e520/src/CSharp/CodeCracker/Usage/NoPrivateReadonlyFieldAnalyzer.cs#L46
This is because the way things were made it would not run always, leaving the old diagnostic hanging longer.
I also changed the compilation not to happen on syntax trees, but on specific syntax nodes (class, structure and field), so the analyzer itself does not need to search, the search is done in a much more faster way by the Roslyn compiler itself, which just calls us when we need. Most of the code is untouched, I did small changes, even though the document was mostly reformatted.

Issue #290, opened 2015-02-27
BUG: Method not used (CC0068) fails to check partial classes


giggio:
If a method is referenced on a partial class, but on a different document (and syntax tree) CC0068 will be generated incorrectly.
Affects C# and VB.

ElemarJR:
How would you like to fix this one. The easy way is make it works only with no partial class. What do you think?

giggio:
@ElemarJR Not necessary. All you need to do is, when checking the method's parent type, if it has the `partial` modifier, get the symbol, and ask for `typeSymbol.DeclaringSyntaxReferences`, then on each type node, continue the verification.

ElemarJR:
I'm back! Is my PR good enough? Could I do that in VB?

giggio:
Great!

PR #296, opened 2015-03-20
Solving bug for csharp impl #290

ElemarJR:
I would like to check if this is enough. So, I will do that in VB.


carloscds:
@ElemarJR Elemar, This code call fixer, but I have a call to method in a partial class:

''''csharp
public partial class foo
    {
        public foo()
        {

        }

        private void test()
        {
        }
    }

    public partial class foo
    {
        public void test2()
        {
            test();
        }
    }
''''

ElemarJR:
I wrote a test with this scenario. Are you sure?

ElemarJR:
I just added another test with the specific scenario proposed by you... passing.

carloscds:
@ElemarJR Ok, but in this case, fixer shouldn´t remove method, right ?

ElemarJR:
@carloscds fixer will never be invoked if the diagnostic does not reports it. Am I right?

giggio:
:+1:

giggio:
C# is done. Now VB! :D

Issue #23, opened 2014-11-17
Remove unused variables


ElemarJR:
Replaces 

````csharp
void Foo()
{
    double a = 10;
    return;
}
````
with 

````csharp
void Foo()
{
    return;
}
````

Category: `Usage`
Diagnostic Id: `CC0056`
Severity: `Info`

@julianosaless will work on it

giggio:
Closed on #299 and #300.

PR #299, opened 2015-03-24
Add remove unused variable #23

julianosaless:
Remove unused variables

### DIFF
--- a/src/CSharp/CodeCracker/Usage/RemoveUnusedVariablesAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Usage/RemoveUnusedVariablesAnalyzer.cs
@@ -1,0 +1,85 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using System.Linq;
+using Microsoft.CodeAnalysis.Text;
+
+namespace CodeCracker.CSharp.Usage
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class RemoveUnusedVariablesAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string Title = "Variable Unused";
+        internal const string MessageFormat = "The variable {0} is never used";
+        const string Description = "When a variable declares and does not use it might bring incorrect conclusions.";
+        internal const string Category = SupportedCategories.Usage;
+        internal static DiagnosticDescriptor Rule = new DiagnosticDescriptor(
+            DiagnosticId.RemoveUnusedVariables.ToDiagnosticId(),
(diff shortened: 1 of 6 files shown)


src/CSharp/CodeCracker/Usage/RemoveUnusedVariablesAnalyzer.cs
```diff
@@ -0,0 +1,85 @@
...
+            if (methodOrConstructor?.Body?.Statements.FirstOrDefault() == null || methodOrConstructor?.Body?.Statements.LastOrDefault() == null) return;
+
+            var dataFlowAnalysis = semanticModel.AnalyzeDataFlow(methodOrConstructor?.Body?.Statements.FirstOrDefault(), methodOrConstructor?.Body?.Statements.LastOrDefault());
+
+            foreach (var variable in variables)
+            {
+                if (!IsUsedVariable(variable, semanticModel, dataFlowAnalysis))
+                {
+                    ReportDiagnostic(context, variable);
+                    return;
```
viniciushana: Wouldn't it just report one diagnostic and exit the method altogether? Perhaps it's better to report all diagnostics without a return statement, letting the foreach loop run for all variables in this context.

julianosaless: @viniciushana  really makes sense, i´ll remove the return :)

viniciushana:
Looks good!

giggio:
I don't like that we are running this analyzer again. The C# compiler already checks for this with `CS0168`: https://msdn.microsoft.com/en-us/library/tf85t6e4%28v=vs.90%29.aspx
You can delete the analyzer and only use the code fix, @julianosaless. Can you fix that?

giggio:
@viniciushana You merged an issue @carloscds was already working on.

carloscds:
@giggio I was looking, but @viniciushana commited.

viniciushana:
Sorry about that, didn't pay attention.

julianosaless:
@giggiomy analyzer is better  kkkkk,  i´ll remove it :) sorry

carloscds:
@julianosaless did you keep fixer ?

julianosaless:
@carloscds i´ll remove only  analyzer.

julianosaless:
Não gostei muito da maneira que a  C# compiler desenvolveu, 
ele possue  2 analyzer, quando uma variável tem valor e quando uma variável não tem, o analyzer que eu desenvolvi engloba as duas possibilidades abaixo: 

CS1688 : The variable 'var' is assigned but its value is never used

public class clz
{
   public static void Main()
   {
      int j = 0;   // CS0168, uncomment the following line
      // j++;
      clx a;       // CS0168, try the following line instead
      // clx a = new clx();
   }
}


CS0219 :  The variable 'variable' is assigned but its value is never used

public class MyClass
{
   public static void Main()
   {
      int a = 0;   // CS0219
   }
}

Porém  no meu Fix eu vou atribuir os dois analyzer para ele: ImmutableArray.Create(new[] { "CS0168", "CS0219" });

giggio:
Closes #23.

PR #303, opened 2015-03-31
String format args

### LINKED ISSUE #116: Check arguments in String.Format
Check if number of arguments passed to `string.Format` is correct. Just analyzer, no codefix.

Category: `Usage`
Severity: `Error`
Diagnostic Id: `CC0056`

giggio:
Closes #116.

### DIFF
--- a/src/CSharp/CodeCracker/Usage/StringFormatArgsAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Usage/StringFormatArgsAnalyzer.cs
@@ -44,17 +44,10 @@ namespace CodeCracker.CSharp.Usage
             if (!argumentList.Arguments[0]?.Expression?.IsKind(SyntaxKind.StringLiteralExpression) ?? false) return;
             if (memberSymbol.ToString() == "string.Format(string, params object[])" && argumentList.Arguments.Skip(1).Any(a => context.SemanticModel.GetTypeInfo(a.Expression).Type.TypeKind == TypeKind.Array)) return;
             var formatLiteral = (LiteralExpressionSyntax)argumentList.Arguments[0].Expression;
-            var format = (string)context.SemanticModel.GetConstantValue(formatLiteral).Value;
-            var formatArgs = Enumerable.Range(1, argumentList.Arguments.Count - 1).Select(i => new object()).ToArray();
-            try
-            {
-                string.Format(format, formatArgs);
-            }
-            catch (FormatException)
-            {
-                var diag = Diagnostic.Create(Rule, invocationExpression.GetLocation());
-                context.ReportDiagnostic(diag);
-            }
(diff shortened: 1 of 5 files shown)


src/CSharp/CodeCracker/Usage/StringFormatArgsAnalyzer.cs
```diff
@@ -0,0 +1,53 @@
...
+            var memberSymbol = context.SemanticModel.GetSymbolInfo(memberExpresion).Symbol;
+            if (memberSymbol == null) return;
+            if (!memberSymbol.ToString().StartsWith("string.Format(string, ")) return;
+            var argumentList = invocationExpression.ArgumentList as ArgumentListSyntax;
+            if (argumentList?.Arguments.Count < 2) return;
+            if (!argumentList.Arguments[0]?.Expression?.IsKind(SyntaxKind.StringLiteralExpression) ?? false) return;
+            if (memberSymbol.ToString() == "string.Format(string, params object[])" && argumentList.Arguments.Skip(1).Any(a => context.SemanticModel.GetTypeInfo(a.Expression).Type.TypeKind == TypeKind.Array)) return;
+            var formatLiteral = (LiteralExpressionSyntax)argumentList.Arguments[0].Expression;
+            var analyzingInterpolation = (InterpolatedStringExpressionSyntax)SyntaxFactory.ParseExpression($"${formatLiteral.Token.Text}");
+            if (analyzingInterpolation.Contents.Count(c => c.IsKind(SyntaxKind.Interpolation)) == argumentList.Arguments.Count - 1) return;
```
viniciushana: While I know it's a matter of personal style and the way that we have to do stuff with Roslyn doesn't help, I'd love to see some of the more complicated validations here extracted to meaningful methods. I think that block full of complex conditions gets quite overwhelming sometimes.

viniciushana:
Overall it looks pretty good, and besides my inline comments, I missed tests with string interpolation. Are they supposed to emit diagnostics? How should this analyzer behave with them? There was code that seemed to ignore interpolated strings but, again, this is only an assumption since there was no test to check for this.

src/CSharp/CodeCracker/Usage/StringFormatArgsAnalyzer.cs
giggio: I usually do that when the complexity gets high, but 16 lines on a method is not that bad. And this code is pretty much self explanatory, IMO.

giggio:
@viniciushana String interpolation test added on line 110 of the test class.

src/CSharp/CodeCracker/Usage/StringFormatArgsAnalyzer.cs
viniciushana: My comment was not much on the line count of the method, instead it's about the complexity of some calls. Anyway, it's more of a personal thing so it doesn't matter so much.

viniciushana:
:+1:

Issue #304, opened 2015-04-05
View lines covered by tests in coverwalls


Ridermansb:
Coverwalls has a feature to show which lines of code are being covered by tests.

Apply this feature to coverwalls.

PR #305, opened 2015-04-05
Send code to coveralls with --full-sources flag

Ridermansb:
ref #304 , Add flag to send full sources

### DIFF
--- a/build.targets.ps1
+++ b/build.targets.ps1
@@ -195,6 +195,6 @@ function RunTestWithCoverage($fullTestDllPaths) {
     Exec { . $reportGeneratorExe -verbosity:Info -reports:$outputXml -targetdir:$coverageReportDir }
     if ($env:COVERALLS_REPO_TOKEN -ne $null) {
         Write-Host -ForegroundColor DarkBlue "Uploading coverage report to Coveralls.io"
-        Exec { . $converallsNetExe --opencover $outputXml }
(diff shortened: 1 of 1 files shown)


giggio:
I don't think this was necessary. This option is not available on coveralls.io nuget:
https://github.com/coveralls-net/coveralls.net/wiki/CI-Integrations#appveyor
Also, if you take two builds:

* https://coveralls.io/builds/2233789
* https://coveralls.io/builds/2269848

The first one did not have the change applied, the second one does. They both show sources. @Ridermansb, why not just send the digest as before? What do we gain from sending full sources?
See coveralls documentation: https://coveralls.zendesk.com/hc/en-us/articles/201774865-API-Introduction

Ridermansb:
@giggio,  [this](https://coveralls.io/builds/2232918/source?filename=src%2FCommon%2FCodeCracker.Common%2FHelpLink.cs) is before full source, and [this](https://coveralls.io/builds/2269785/source?filename=src%2FCommon%2FCodeCracker.Common%2FHelpLink.cs) is after full source.

I believe that sent the complete code prevents problems with reference to the repository since the entire code is in coveralls.

The coveralls-net have this option, [see](https://github.com/coveralls-net/coveralls.net/blob/master/coveralls.net/CommandLineOptions.cs#L58)

The default was not send.

giggio:
Weird thing, that first build I mentioned had the files. Anyway, it seems to be working better, as you showed it. Thanks!

PR #312, opened 2015-04-16
Add compute value of an expression analyzer

### LINKED ISSUE #117: Compute value of an expression and replaces it whenever it's possible
`Console.WriteLine(1 + 1)` becomes `Console.WriteLine(2)`

This is `Hidden` and `Refactoring`.
Diagnostic Id: `CC0082`

giggio:
Closes #117

### DIFF
--- a/src/CSharp/CodeCracker/Refactoring/ComputeExpressionCodeFixProvider.cs
+++ b/src/CSharp/CodeCracker/Refactoring/ComputeExpressionCodeFixProvider.cs
@@ -1,0 +1,56 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using System.Collections.Immutable;
+using System.Composition;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace CodeCracker.CSharp.Refactoring
+{
+
+    [ExportCodeFixProvider("CodeCrackerComputeExpressionCodeFixProvider", LanguageNames.CSharp), Shared]
+    public class ComputeExpressionCodeFixProvider : CodeFixProvider
+    {
+
+        public sealed override ImmutableArray<string> FixableDiagnosticIds =>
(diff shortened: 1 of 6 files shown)


src/CSharp/CodeCracker/Refactoring/ComputeExpressionCodeFixProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+            if (newRoot == null) return null;
+            var newDocument = document.WithSyntaxRoot(newRoot);
+            return newDocument;
+        }
+
+        internal static SyntaxNode ComputeExpression(SyntaxNode nodeToReplace, BinaryExpressionSyntax expression, SyntaxNode root, SemanticModel semanticModel)
+        {
+            var result = semanticModel.GetConstantValue(expression);
+            if (!result.HasValue) return null;
+            var newExpression = SyntaxFactory.ParseExpression(result.Value.ToString());
```
carloscds: @giggio This line is produce an error in value because of culture. Result value is 2.4 and after result.Value.ToString() is 2,4. and two tests fail.

giggio: @carloscds I have updated it, but I can't test it, because my system is in en-US. Can you please verify?

carloscds: @giggio Perfect now!

giggio: Nice! @carloscds @ElemarJR Do you have any idea on how we could add a unit test to verify this?

carloscds: @giggio @ElemarJR Do you talk about check culture ?

giggio: @carloscds @ElemarJR yes.

carloscds: @giggio @ElemarJR It's possible to create a test that check culture, but in this case we will force user to change your culture to en-us or put an alert (inconclusive test) about this ? Maybe put something in documentation about tests using numeric conversions. What do you think ?

giggio: @carloscds @ElemarJR I am not sure how to solve this. I created issue #313 so we can discuss there.

PR #320, opened 2015-04-21
Allow to change culture in a test

### LINKED ISSUE #313: Create infrastructure for testing multi culture values
On #312 a test failed when we tested on a pt-BR version of VS. We need a way to verify these types of issues on our unit tests.
We need to discuss what would be the best approach to do this.

giggio:
Closes #313, which describes the issue in more detail.

It adds the `ChangeCulture` class which changes the current culture
during a test. It enables testing on specific cultures, that were harder
to test before. It is a simple wrap on the current culture property on
the current thread.
I also added on every await on the base and helper test classes a
`.ConfigureAwait(true)` so that every test always returns on the same
thread as it started. If we did not do this then the thread that
continues a test may be different from the one that started it, causing
a problem on the current culture setting.


carloscds:
@giggio I didn´t understand relationship between culture problem and configureawait()

PR #329, opened 2015-04-28
broken build

sawilde:
I couldn't get the build to work locally until I made the following changes

(the packages.config changes were actually made by VS2015 as it restored the packages)

### DIFF
--- a/.nuget/packages.config
+++ b/.nuget/packages.config
@@ -1,7 +1,9 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
+  <package id="coveralls.io" version="1.3.2" />
   <package id="NuGet.CommandLine" version="2.8.2" />
   <package id="OpenCover" version="4.5.3809-rc94" />
   <package id="psake" version="4.4.1" />
   <package id="ReportGenerator" version="2.1.3.0" />
+  <package id="xunit.runner.console" version="2.0.0" />
 </packages>
--- a/build.targets.ps1
+++ b/build.targets.ps1
@@ -149,7 +149,7 @@ function RestorePkgs($sln) {
         Write-Host "File $sln does not exist, exiting." -ForegroundColor DarkRed
         return
     }
-    nuget restore $sln
+    exec { . $nugetExe restore $sln }
 }
 
 function TestPath($paths) {


carloscds:
@sawilde What version of VS are you using ?

sawilde:
VS2015 CTP6 - however it was on the command line that nuget wouldn't run until I made the chnage to build.targets.ps1

carloscds:
@sawilde What error you receive ? Because I'm running CTP 6 too, without errors

sawilde:
I am prepping a new version of opencover and was checking that the new Mono.Cecil handles the issues that code-cracker raised with the new PDB format and Mono.Cecil

No errors in CTP6 itself - actually the command line

.\Build.ps1 Build

sawilde:
````
Windows PowerShell
Copyright (C) 2014 Microsoft Corporation. All rights reserved.

PS C:\Users\Shaun> cd C:\Projects\external\code-cracker\
PS C:\Projects\external\code-cracker> .\build.ps1
psake version 4.4.1
Copyright (c) 2010-2014 James Kovacs & Contributors

-------------------------[Restore]-------------------------
Restoring C:\Projects\external\code-cracker\CodeCracker.CSharp.sln...
Error: 4/28/2015 10:50:23 PM:
At C:\Projects\external\code-cracker\build.targets.ps1:152 char:5 +     nuget restore $sln +     ~~~~~ [<<==>>] Excepti
on: The term 'nuget' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the s
pelling of the name, or if a path was included, verify that the path is correct and try again.
````

carloscds:
@sawilde Can you try to build using Visual Studio IDE ?

sawilde:
Why? I have no problem with the Visual Studio build.

The problem is with the command line build; as shown above.

carloscds:
@sawilde Even after your changes, simply executing build.ps1 can't restore packages, because you need Nuget Command Line. How do you this test after change ?

carloscds:
@sawilde if you download nuget.exe and restore packages, you can execute build.

carloscds:
@sawilde Your change in build.targets is perfect, but don't need change packages.config.

sawilde:
why not? When I restored packages via VS2015 it automatically updated that file, which when you look at those packages is the right place for them to be registered as they are tools.

giggio:
Both changes make sense to me. It's a weird thing VS never asked me to update packages.config, but I believe this could happen. If we change that, then, the best thing would be to remove it from the Test.Common project where I believe those deps referenced. (I'm away from a computer now)

sharwell:
BTW, if you ever want to make a change where the build works even if the user doesn't have **NuGet.exe** downloaded, feel free to use [this block of code](https://github.com/rackerlabs/dotnet-collections/blob/a970c03a3ac8c9676390343f0381e7d7e491b52b/build/build.ps1#L31-L39) in the build script.

sawilde:
OT: How did you get that coveralls thing to report back?

giggio:
@sawilde I don't know. It just started doing it.

Issue #198, opened 2015-01-05
Test NameOfCodeFixProvider with keyword


sharwell:
Two tests should be added in support of C# code using the `@identifier` syntax:

1. Test with a parameter name which is a C# keyword (e.g. `@object`)
2. Test with a parameter name which is *not* a C# keyword (e.g. `@obj`)

@baks is working on it.

baks:
Looks easy for somebody new like me :)
@giggio  I take this one

giggio:
Closed on #331.

PR #331, opened 2015-05-05
Test with verbatim identifier NameOfCodeFixProvider (#198)

baks:
Regarding to #198 I added tests to NameOfCodeFixProvider with verbatim identifier.

It turns out that NameOf doesn't work correctly with verbatim identifier.

NameOfCodeFixProvider changed this code:
```csharp
        void Foo(string @for)
        {
            string whatever = "for";
        }
```
to following:
```csharp
        void Foo(string @for)
        {
            string whatever = nameof(for);
        }
```

So I changed NameOfCode(Analyzer/FixProvider) to take nameof argument from identifier that the string literal refers and now NameOf seems working correctly.


giggio:
@baks @carloscds It looks good to me. I did not run it, but it looks fine.
And @baks, nice catch on the verbatim identifier!

giggio:
We are still missing the vb fix for this same issue. But I think @baks will do it in another issue.

Issue #319, opened 2015-04-21
Bug: CC0022 being raised on using when not assigned


giggio:
This is incorrectly raising `CC0022`:

````csharp
using (new SomeDisposable()) { }
````

It shouldn't, as the object created from `SomeDisposable` will disposed. The analyzer works as expected when the value is assigned, like so:

````csharp
using (var d = new SomeDisposable()) { }
````

PR #336, opened 2015-05-11
Add analyzer/fix for redudant field initializers

### LINKED ISSUE #9: Supress assignment of default value to field/property declarations
In this cases:

````csharp
private class C
{
    private bool aBool = false;
    private int i = 0;
    private DateTime d = DateTime.MinValue;
    private MyClass mc = null;
    private MyStruct ms = default(MyStruct);
    //etc...
}
````

Becomes:

````csharp
private class C
{
    private bool aBool;
    private int;
    private DateTime d;
    private MyClass mc;
    private MyStruct ms;
    //etc...
}
````

This analyzer applies to all fields which are not marked `const`. For each field *f*, let `T` be the declared type of *f*. If *f* is initialized with the value `default(T)`, then report this diagnostic. The following additional cases should be handled:

1. If `T` is a reference type or a nullable value type, then treat `null` the same as `default(T)`. Note: this means you can always report this diagnostic for `f = null`, regardless of the actual type `T`.
2. If `T` is a numeric primitive type, then treat `0` (or equivalent double or float) value as `default(T)`.
3. If `T` is an enum type, then treat `0` as equivalent to `default(T)`. Note: do NOT treat an enum member this way, even if the value of the enum member is 0.
4. If `T` is `IntPtr`, then treat `IntPtr.Zero` equivalent to `default(T)`.
5. If `T` is `UIntPtr`, then treat `UIntPtr.Zero` equivalent to `default(T)`.
6. If `T` is `DateTime`, then treat `DateTime.MinValue` equivalent to `default(T)`.

Should use `WellKnownDiagnosticTags.Unnecessary`.

Severity: `Info`.
Diagnostic id is `CC0034`.
Category: `Usage`

giggio:
closes #9

PR created from Greece during my vacations! :) :sunny:

### DIFF
--- a/src/CSharp/CodeCracker/CodeCracker.csproj
+++ b/src/CSharp/CodeCracker/CodeCracker.csproj
@@ -102,6 +102,7 @@
     <Compile Include="Style\StringFormatAnalyzer.cs" />
     <Compile Include="Usage\DisposableFieldNotDisposedAnalyzer.cs" />
     <Compile Include="Usage\DisposableVariableNotDisposedCodeFixProvider.cs" />
+    <Compile Include="Usage\RedundantFieldAssignmentCodeFixProvider.cs" />
     <Compile Include="Usage\RemoveUnreachableCodeFixAllProvider.cs" />
     <Compile Include="Usage\DisposableVariableNotDisposedFixAllProvider.cs" />
     <Compile Include="Usage\RemovePrivateMethodNeverUsedAnalyzer.cs" />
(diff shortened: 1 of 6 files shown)


carloscds:
@giggio This test is breaking: public async Task ComputeExpression(string original, string fix) =>
            await VerifyCSharpFixAsync(original.WrapInCSharpMethod(), fix.WrapInCSharpMethod());
image

giggio:
Hm... same problem of globalization again, I will check it out.

giggio:
@carloscds This test `ComputeExpression` is not from this Pull Request. This PR is about redundant fields, and I did not change the CompureExpression files, as you can see from the commits.
You had already caught this error on #312, and I had already fixed it. We even worked on #313 because of it, remember?
Are you up to date with the ´master` branch?

carloscds:
@giggio Yes, I know this. My suggestion is open a bug to solve this.

giggio:
@carloscds I can't repro. As I told you, I already fixed that bug. See this line:
https://github.com/code-cracker/code-cracker/blob/master/src/CSharp/CodeCracker/Refactoring/ComputeExpressionCodeFixProvider.cs#L51
Is this PR ok?

carloscds:
@giggio I remember this, but now in RC I have same bug when I run tests.

giggio:
Ok @carloscds, bug #319 opened for that unrelated issue.

How about this specific PR? Is it ok?

Issue #335, opened 2015-05-10
Incorrect string.Format usage


giggio:
This can be identified as a flawed string but isn't:

````CSharp
string.Format(@"{1}", 1);
````

There is no code fix.

We could expand on `CC0056` or not. Let's discuss it. If we decide to create a new analyzer, it should follow `CC0056` details (Error, Usage).

Here is the current analyzer for `CC0056`: https://github.com/code-cracker/code-cracker/blob/master/src/CSharp/CodeCracker/Usage/StringFormatArgsAnalyzer.cs

@thorgeirk11 is working on it.

giggio:
Looking at the current implementation it shouldn't be a problem to expand `StringFormatArgsAnalyzer` to make it solve this problem. @thorgeirk11 you brought it up on #330, and fixed it later, what do you think? @code-cracker/owners what do you think?

carloscds:
@giggio Why you close another bug about this ? Solve only this problem not solve it.

carloscds:
I think that it´s not a problem for analyzer, because it´s a basic error. But if someone will want to fix it, go ahead!

thorgeirk11:
I would like to try to extend CC0056 to also include this error, if that is ok ? 
What would the error message for this scenario be and would it be the same diagnostic ID ?

PR #337, opened 2015-05-12
CC0056 Identify invalid format strings

thorgeirk11:
Discussed in #335, I have extended the StringFormatArgsAnalyzer to also identify invalid format strings.

### DIFF
--- a/src/CSharp/CodeCracker/Usage/StringFormatArgsAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Usage/StringFormatArgsAnalyzer.cs
@@ -12,21 +12,32 @@ namespace CodeCracker.CSharp.Usage
     public class StringFormatArgsAnalyzer : DiagnosticAnalyzer
     {
         internal const string Title = "Incorrect String.Format usage";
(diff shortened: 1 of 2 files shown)


giggio:
@thorgeirk11  I just overviewed the code, it looks good, and tests pass, and coverage increased (yey!). I will have to try it and if everything is ok I will merge.

thorgeirk11:
Awesome.

thorgeirk11:
Woops, did not mean that to happen

giggio:
I have merged it but as the update works perfectly. I am just not sure that we should be using 2 different messages with the same diagnostic id. I am thinking we should use a different id.
@code-cracker/owners please see that we have two rules with the same diagnostic id and different messages:
https://github.com/code-cracker/code-cracker/blob/master/src/CSharp/CodeCracker/Usage/StringFormatArgsAnalyzer.cs#L20-L38
**Note:** Do you think that is ok or should we create a new id? I am leaning towards creating a new one.
There is no fix for the diagnostic generated by `StringFormatArgsAnalyzer`, so the impact would be minimal.

carloscds:
@giggio About this analyzer in particular, I guess that is not a problem.

Issue #263, opened 2015-02-07
Expand NameOf Analyzer to work with any identifier in scope


giggio:
Right now `CC0021` is only working with function parameters. It should work with any analyzer in Scope, so class names, functions names, etc.
Code fix does not need to be updated, it should work as is.

jwooley:
Should be updated in c# and VB

baks:
I recently worked on #198 concerning NameOfAnalyzer so if possible I could take this one.

giggio:
@baks Ok, go ahead, it is yours. If you find the problem from #198 problem here just update it as well.
I invited you to the collaborators team so I can assign it to you.

baks:
@giggio Thank you for invitation

giggio:
VB is pending, issue #341 was opened so we don't forget.

PR #340, opened 2015-05-15
C# CC0021 analyzer expand scope

baks:
Fixes C# part of #263 

Changed C# NameOfAnalyzer to work with any scope.
Now NameOfAnalyzer first check if string literal is in method/constructor/indexer and tries to find match in parameters list.
If that fails semantic model is used to lookup symbols/namespaces.


giggio:
@baks It is merged. I had to update a few things. Check out the 2 commits to see what I have done.

baks:
@giggio, thanks! 
I checked the changes that you've done and must admit that you simplify a lot NameOfTests :+1: 
And next time I'll better follow code-cracker code style

PR #342, opened 2015-05-15
Use StringEmpty

### LINKED ISSUE #120: Use String.Empty instead ""
````csharp
var a = "";
````

Becomes 

````csharp
var a = string.Empty;
````

`Style`, `Info`

Diagnostic Id: `CC0084`
Severity: Hidden
Category: Style

carloscds:
Close #120

### DIFF
--- a/src/CSharp/CodeCracker/CodeCracker.csproj
+++ b/src/CSharp/CodeCracker/CodeCracker.csproj
@@ -57,6 +57,8 @@
     <Compile Include="Refactoring\ParameterRefactoryCodeFixProvider.cs" />
     <Compile Include="Reliability\UseConfigureAwaitFalseAnalyzer.cs" />
     <Compile Include="Reliability\UseConfigureAwaitFalseCodeFixProvider.cs" />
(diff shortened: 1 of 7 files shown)


giggio:
@carloscds You are analyzing only variable declarations, you should instead be analyzing string literals.
For example, this will not currently raise a diagnostic:

````csharp
void Foo(string s) {}
void Bar()
{
    Foo("");
}
````

But it should.

giggio:
@carloscds I am also not sure if this is an `Info` or a `Refactoring`. This is a such a matter of taste. What do you think. Also, @ElemarJR, you proposed #120, what do you think?

carloscds:
@giggio Look at the test wrote by @ElemarJR. Specification is only about variables.

giggio:
@carloscds it is just an example, it does not mean we should not solve the more generic problem. @ElemarJR do you agree? Carlos, it should not be too hard to change it to a more generic solution.

carloscds:
@giggio Ok. it's make sense. I will implement.

giggio:
@carloscds There were some errors on the fix all, and I also found a problem when using string.empty on attributes, so I excluded them.

PR #346, opened 2015-05-17
Add UseStaticRegexIsMatch Analyzer, closes #297

### LINKED ISSUE #297: Regex performance: use static Regex.IsMatch
We could diagnose code using Regex in a way that gives poor performances.

Creating a new Regex every time will give you a bad performance:

```C#
public static bool IsValidCurrency(string currencyValue)
{
  string pattern = @"\p{Sc}+\s*\d+";
  Regex currencyRegex = new Regex(pattern);
  return currencyRegex.IsMatch(currencyValue);
}
```

If this method is called multiple times, you should be using the [static `Regex.IsMatch`](https://msdn.microsoft.com/en-us/library/sdx2bds0%28v=vs.110%29.aspx) method, as that can be [cached](https://msdn.microsoft.com/en-us/library/gg578045(v=vs.110).aspx) and won't be interpreted every time you call the method, it will give you better performance.
This is quite common on email address validation during registration, or validating user input (like the currency example above or Zip codes).

Code fix:
```C#
public static bool IsValidCurrency(string currencyValue)
{
  string pattern = @"\p{Sc}+\s*\d+";
  return Regex.IsMatch(currencyValue, pattern); 
}
```
I created a [project](https://github.com/akamud/RegexPerformanceExamples/blob/master/RegexPerformanceExamples/Program.cs) showing the difference in performance, testing this case with only 10 runs shows that instance method is almost 3 times slower.
```
Instance took 0.000639 seconds
Static took   0.000240 seconds
```

Running this 100,000 times gives these results:
```
Instance took 0.3786505 seconds
Static took   0.0717608 seconds
```

I'm sure there are [more cases](http://blogs.msdn.com/b/bclteam/archive/2010/06/25/optimizing-regular-expression-performance-part-i-working-with-the-regex-class-and-regex-objects.aspx) of improving Regex performance that we could diagnose, but this one is pretty straightforward. What do you think?

Diagnostic Id: `CC0081`
Severity: `Info`
Style: `Performance`

@akamud is working on it.

akamud:
As suggested, it provides 3 possible code fix: use static, use compiled, use both.

### DIFF
--- a/src/CSharp/CodeCracker/CodeCracker.csproj
+++ b/src/CSharp/CodeCracker/CodeCracker.csproj
@@ -35,8 +35,10 @@
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="Extensions\CSharpAnalyzerExtensions.cs" />
+    <Compile Include="Performance\UseStaticRegexIsMatchAnalyzer.cs" />
     <Compile Include="Performance\StringBuilderInLoopCodeFixProvider.cs" />
     <Compile Include="Performance\StringBuilderInLoopAnalyzer.cs" />
+    <Compile Include="Performance\UseStaticRegexIsMatchCodeFixProvider.cs" />
     <Compile Include="Refactoring\AddBracesToSwitchSectionsAnalyzer.cs" />
     <Compile Include="Refactoring\AddBracesToSwitchSectionsCodeFix.cs" />
     <Compile Include="Refactoring\AllowMembersOrderingAnalyzer.cs" />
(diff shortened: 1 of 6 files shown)


giggio:
@akamud I have refactored a little, see 34c9ca4 if you would like to see what changed. I mainly reorganized the code and changed the code fix way of registering the fix to be more efficient. This is something that we will have to do to the other fixes in the project as well, as we recently found out that we can be more efficient.

giggio:
@akamud  BTW, nice job, it worked perfectly. You are more than welcome to pick up another one to help, or suggest a new one as you did before.

akamud:
Thanks @giggio, I'll take a look at your commit to understand what are the changes related to efficiency for my next contribution :)

Is there a label identifying all the other fixes that still need to be changed? I could help with that.

giggio:
@akamud yes. :)
If you want to help with some, just do it and send a PR. You don't need to do the whole project, it is a lot. I will also do the C# ones, but I have not yet started. Just comment on the issue if you want to help.

PR #351, opened 2015-05-21
Update RemovePrivateMethodNeverUsed to work with generic methods

### LINKED ISSUE #343: Bug on CC0068 (RemovePrivateMethodNeverUsed) with generic methods
Right now it seems that generic methods are breaking `CC0068`.
This method:

````csharp
private static SymbolAnalysisContext GetSymbolAnalysisContext<T>(string code, string fileName = "a.cs") where T : SyntaxNode
````

(Also here: https://github.com/code-cracker/code-cracker/blob/master/test/CSharp/CodeCracker.Test/GeneratedCodeAnalysisExtensionsTests.cs#L305)

Is raising a `CC0068` when it shouldn't.

giggio:
The first commit is better to analyze the PR. The 2nd one is to update from tabs to spaces, the file was incorrectly formatted.
Closes #343.


giggio:
@carloscds I have rebased this to work with the new `HEAD`. If you have checked it out you need to delete your local branch and do it again.

Issue #354, opened 2015-05-22
Use "" instead of String.Empty


giggio:
This is the opposite of #120.

````csharp
var a = string.Empty;
````

Becomes 

````csharp
var a = "";
````

Diagnostic Id: `CC0088`
Severity: Hidden
Category: Style

@kalkie is working on it.

kalkie:
Hi giggio, can I try to solve this one?

giggio:
@kalkie yes, go ahead. I'll also add you to the collaborators team and assign it to you.

PR #362, opened 2015-05-27
#354 use literal empty string instead of string.empty

kalkie:
This implements #354 Use "" instead of String.Empty. Let me know if there are any remarks.


carloscds:
@kalkie Good job!

kalkie:
Thanks @carloscds

giggio:
Diagnostic id was 88. I just fixed it in 1314b17. Also, the code fix attribute was different, the format was recently updated by Microsoft and we followed.

Issue #347, opened 2015-05-18
Update all existing code fixes that are doing too much work on RegisterCodeFixesAsync (C#)


giggio:
Right now most of the code fix providers are doing more work than they should on the `RegisterCodeFixesAsync` method.
They should delegate most, if not all, of the work to the code actions.

See the expected way to do this on:
https://github.com/code-cracker/code-cracker/blob/master/src/CSharp/CodeCracker/Performance/UseStaticRegexIsMatchCodeFixProvider.cs#L22

And how the correction was applied on 34c9ca4a819d180.

This is for C# only. For VB view issue #348.

giggio:
I did most of the work, but we are still missing the `Style` and `Usage` code fixes.

giggio:
I am doing the Style ones now.

giggio:
I am doing Usage now.

PR #374, opened 2015-06-09
CS0051 code fix

### LINKED ISSUE #321: Make accessibility consistent (code fix for CS0051)
If a class references another class that is less accessible you will get `CS0051`. For example

````csharp
public class Dependent
{
    public Dependent(DependendedUpon d)
    {
    }
}
internal class DependendedUpon
{
}
````

This will raise `CS0051` on the first caracter of the constructor identifier (the first letter `D`).

The code fix should simply change the accessibility of the less accessible class. It would change the above code to (showing only the changed code, the first class stays the same, note the `public` modifier):

````csharp
public class DependendedUpon
{
}
````

There is no need for diagnostic id (we are using the `CS` one, category or visibility, as this is only a code fix provider.

### PR DESCRIPTION
**Pull request by baks**
Fixes #321 

Code fix for CS0051 and also initial structure for other errors related with inconsistent accessibility.

InconsistentAccesibilityProviders should return InconsistentAccessibilityInfo with TypeSyntax, new accessibility modifiers and code action message. With that information code fix is able to search type symbol, locate proper document in solution and change accessibility modifiers for that type.

@giggio, let  me know what do you think

### CODE REVIEW

File: src/CSharp/CodeCracker/Design/InconsistentAccessibility/InconsistentAccessibilityCodeFixProvider.cs [hunk 1]
```diff
@@ -0,0 +1,159 @@
...
+                if(typeSymbol != null)
+                {
+                    var document = context.Document.Project.Solution.GetDocument(typeSymbol.Locations[0].SourceTree);
+
+                    if(document != null)
+                    {
+                        context.RegisterCodeFix(CodeAction.Create(info.CodeActionMessage, c => ChangeTypeAccessibilityInDocument(document, info.NewAccessibilityModifiers, typeSymbol.Locations[0].SourceSpan, c)), diagnostic);
+                    }
+                }
+            }
```
giggio: `RegisterCodeFixesAsync` is doing too much work. See #347 where we are already changing older analyzers to perform better. It seems there is a way to delegate all this work to the Lambda in `CodeAction.Create`. Please update the provider to match that pattern, unless we really need to do some work on this method.

baks: Ok, I'll change that

baks: `RegisterCodeFixesAsync` checks whether type declaration exists in solution.
I think we can't move this to `CodeAction.Create` because if the type declaration is only in the metadata we shouldn't propose code fix at all.

Actually we can move part that finds Document and in `RegisterCodeFixesAsync` only see wheter any `Location` is in source. Agree?

giggio:
@baks The type has to exist on the solution, otherwise you couldn't even be referencing it, right? For example:

File Person.cs in your solution:
````csharp
class Person
{
    void WalkTheDog(Dog dog)
    {
    }
}
````

File Dog.cs outside of your solution, coming to you as a reference from a dll:
````csharp
class Dog
{
}
````

This will actually raise `CS0122`, not `CS0051`.

This scenario would work even on the same solution, with files being on different projects.

I guess `CS0051` is only raised when a method can reference the type, but not expose it outside as a parameter. If the type cannot be referenced, then we get `CS0122`.

BTW, if you can rebase or merge the changes from master it would be great. I updated some testing infrastructure.

**Comment by baks** [comment]
Ok, I'll merge with master.

@giggio, If you create class like that in other assembly:
```csharp
public class Dog
{
    protected class Walker
    {
    }
}
```
and you use it in your solution:
```csharp
public class EnhancedDog
{
    public EnhancedDog(Walker walker)
    {
    }
}
```
This will raise `CS0051`.
Or if you use `InternalsVisibleToAssembly()` in example that you provided it should also raise `CS0051` (I'm not sure about this on 100% percent, but I'll check this when I get home).

giggio:
Hm... you are right!
We do need that check, then. If you can move all the other code to the other methods, away from `RegisterCodeFixesAsync` method, it would be fine.

giggio:
BTW, another possible code fix would be to make the `EnhancedDog` constructor `protected`, right? Let's finish this one, then we can think about it later.

baks:
@giggio , exactly :+1: 

This is the same as other fix I proposed in #321

baks:
@giggio I merged with master and included PR comments about `RegisterCodeFixesAsync`

src/CSharp/CodeCracker/Design/InconsistentAccessibility/InconsistentAccessibilityCodeFixProvider.cs
```diff
@@ -0,0 +1,169 @@
...
+        private static async Task<Location> FindTypeLocationInSourceCodeAsync(Document document, TypeSyntax type, CancellationToken cancellationToken)
+        {
+            var result = Location.None;
+
+            var semanticModel = await document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+            var typeSymbol = semanticModel.GetSymbolInfo(type, cancellationToken).Symbol;
+
+            if(typeSymbol != null)
+            {
+                var locationInSource = typeSymbol.Locations.FirstOrDefault(location => location.IsInSource);
```
giggio: You may find more than one location for a type if the type is declared `partial`. I suggest you return a `Task<Location[]>` or something similar.

src/CSharp/CodeCracker/Design/InconsistentAccessibility/InconsistentAccessibilityCodeFixProvider.cs
```diff
@@ -0,0 +1,169 @@
...
+
+        private static async Task<Document> ChangeTypeAccessibilityAsync(Solution solution, SyntaxTokenList newAccessibilityModifiers, Location typeLocation, CancellationToken cancellationToken)
+        {
+            var document = solution.GetDocument(typeLocation.SourceTree);
+            var syntaxRoot = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+
+            var result = document;
+
+            var declaration = syntaxRoot.FindNode(typeLocation.SourceSpan) as MemberDeclarationSyntax;
+            if (declaration != null)
```
giggio: When we get to the code fix provider we have to be sure of what syntax we are dealing with. There should not be any situation where `SyntaxRoot.FindNode(typeLocation.SourceSpan)` is not a `MemberDeclarationSyntax`.
I suggest you change this to `(MemberDeclarationSyntax)SyntaxRoot.FindNode(typeLocation.SourceSpan)` and remove the null check.

src/CSharp/CodeCracker/Design/InconsistentAccessibility/InconsistentAccessibilityInMethodParameter.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        private static string ExtractParameterTypeFromErrorMessage(string compilerErrorMessage)
+        {
+            const int parameterTypeStartShift = 52;
+
+            var parameterTypeNameStart = compilerErrorMessage.IndexOf(InconsistentAccessibilityCodeFixProvider.InconsistentAccessibilityInMethodParameterCompilerErrorNumber, StringComparison.Ordinal) + parameterTypeStartShift;
+            var parameterTypeNameLength = compilerErrorMessage.IndexOf('\'', parameterTypeNameStart) - parameterTypeNameStart;
+            return compilerErrorMessage.Substring(parameterTypeNameStart, parameterTypeNameLength);
+        }
```
giggio: You should not extract the parameter this way. We will have people that have different cultures in VS and the message will appear in different positions.
Try to change it to something like this:

````csharp
System.Text.RegularExpressions.Regex.Match(
    Microsoft.CodeAnalysis.CSharp.CSharpDiagnosticFormatter.Instance.Format(
        diagnostic, System.Globalization.CultureInfo.InvariantCulture),
        diagnostic.Descriptor.MessageFormat.ToString(System.Globalization.CultureInfo.InvariantCulture).Replace("{0}", "(.*)").Replace("{1}", "(.*)")
).Groups[1].Value
````

baks: I'll change that to way that you proposed.

src/CSharp/CodeCracker/Design/InconsistentAccessibility/InconsistentAccessibilityCodeFixProvider.cs
baks: That's right. I'll also add test for `partial` declaration

src/CSharp/CodeCracker/Design/InconsistentAccessibility/InconsistentAccessibilityCodeFixProvider.cs
baks: :+1:

baks:
@giggio , I changed code fix provider to handle partial classess and also changed extracting parameter type to be culture agnostic with a way that you provided

giggio:
@baks The code is fine, but I found an inconsistent behavior on VS. I filed a bug here on the Roslyn repo: https://github.com/dotnet/roslyn/issues/3496
Because of that I think I will revert to a specific message until this is fixed. Tests pass but no code fix is offered at VS.

giggio:
@baks See 9674315 where I updated the method.

PR #376, opened 2015-06-12
Make method static

### LINKED ISSUE #364: Make method static if possible
If method in class is not using any instance fields it can be made static.

````csharp
class Test 
{
    public char GetThirdCharacterFromString(string str)
    {
        return str[2];
    }
}
class Referencing
{
    void Foo()
    {
        var t = new Test();
        t.GetThirdCharacterFromString("");
    }
}
````

becomes

````csharp
public class Test 
{
    public static char GetThirdCharacterFromString(string str)
    {
        return str[2];
    }
    void Foo()
    {
        var t = new Test();
        Test.GetThirdCharacterFromString("");
    }
}
````

A non static method can be turned static, but only if it is not accessing `this`.
Also, the code fix would have to find all references and change them to static invocations. If the method is public, on the whole solution, if it is internal, on the project, etc. The current roslyn infrastructure takes care of finding those references for us.

Diagnostic id: `CC0091`
Category: `Design`
Severity: `Info`

### PR DESCRIPTION
**Pull request by giggio**
closes #364

### DIFF
--- a/src/CSharp/CodeCracker/Design/MakeMethodStaticAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Design/MakeMethodStaticAnalyzer.cs
@@ -1,0 +1,63 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using System.Linq;
+using System.Collections.Immutable;
+
+namespace CodeCracker.CSharp.Design
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class MakeMethodStaticAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string Title = "Use static method";
+        internal const string MessageFormat = "Make '{0}' method static.";
+        internal const string Category = SupportedCategories.Design;
+        const string Description = "If the method is not referencing any instance variable and if you are " +
+            "not creating a virtual, abstract, new or partial method, and if it is not a method override, " +
+            "your instance method may be changed to a static method.";
+
+        internal static DiagnosticDescriptor Rule = new DiagnosticDescriptor(
+            DiagnosticId.MakeMethodStatic.ToDiagnosticId(),
+            Title,
+            MessageFormat,
+            Category,
+            DiagnosticSeverity.Warning,
+            true,
+            description: Description,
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/CSharp/CodeCracker/Design/MakeMethodStaticAnalyzer.cs [hunk 1]
```diff
@@ -0,0 +1,55 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context) =>
+            context.RegisterSyntaxNodeAction(LanguageVersion.CSharp6, AnalyzeMethod, SyntaxKind.MethodDeclaration);
+
+        private void AnalyzeMethod(SyntaxNodeAnalysisContext context)
+        {
+            if (context.IsGenerated()) return;
+            var method = (MethodDeclarationSyntax)context.Node;
+            if (method.Modifiers.Any(SyntaxKind.StaticKeyword)) return;
```
baks: The analyzer should also immediately return if there are any of this method modifiers:
* `virtual`
* `override`
* `new`

giggio: @baks makes sense. I will updated it.

giggio: @baks Done. Also added `abstract`.

src/CSharp/CodeCracker/Design/MakeMethodStaticAnalyzer.cs
```diff
@@ -0,0 +1,61 @@
...
+
+namespace CodeCracker.CSharp.Design
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class MakeMethodStaticAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string Title = "Use static method";
+        internal const string MessageFormat = "Make '{0}' method static.";
+        internal const string Category = SupportedCategories.Design;
+        const string Description = "If you are not referencing any instance variable your method may become static. That will require";
```
baks: It looks like something is missing in `Description` string or am I wrong? :question:

giggio: @baks  I think I forgot to finish it. I updated and rebased everything.

carloscds:
@giggio Giovanni, can you update your PR for merge ?

giggio:
@carloscds done!

Issue #381, opened 2015-06-14
Make accessibility consistent (code fix for CS0050 to CS0061)


giggio:
**This issue is an umbrella issue. We will only close it when every sub issue is closed.**

Sub issues:

- [x] [`CS0050`](https://msdn.microsoft.com/en-us/library/be25s2h2.aspx) (#387)
- [x] [`CS0051`](https://msdn.microsoft.com/en-us/library/4sscdk02.aspx) (#321)
- [x] [`CS0052`](https://msdn.microsoft.com/en-us/library/997yz02a.aspx) (#391)
- [x] [`CS0053`](https://msdn.microsoft.com/en-us/library/532wtsbc%28v=vs.90%29.aspx) (#391)
- [x] [`CS0054`](https://msdn.microsoft.com/en-us/library/0efy9xb8%28v=vs.90%29.aspx) (#391)
- [x] [`CS0055`](https://msdn.microsoft.com/en-us/library/6b0x5w9f%28v=vs.90%29.aspx) (#391)
- [x] [`CS0056`](https://msdn.microsoft.com/en-us/library/7zdw9dt0%28v=vs.90%29.aspx) (#532)
- [x] [`CS0057`](https://msdn.microsoft.com/en-us/library/1he6k2zs%28v=vs.90%29.aspx) (#532)
- [x] [`CS0058`](https://msdn.microsoft.com/en-us/library/9c065z8s%28v=vs.90%29.aspx) (#558)
- [x] [`CS0059`](https://msdn.microsoft.com/en-us/library/20f51zz7%28v=vs.90%29.aspx) (#558)
- [x] [`CS0060`](https://msdn.microsoft.com/en-us/library/5732hs83%28v=vs.90%29.aspx)
- [x] [`CS0061`](https://msdn.microsoft.com/en-us/library/7f82xxf1%28v=vs.90%29.aspx)

baks:
So, I can start work on `CS0050`. Should I create new issue for 'CS0050' with details of how code fix provider should work?

giggio:
@baks I don't think we need it. Just go ahead.

baks:
`CS0050` is done, so I start doing `CS0052`

giggio:
Half is done! :)

baks:
@giggio 
Sorry, it takes so long time :cry: , but now I'm able to finish it.

giggio:
Cool!

giggio:
And we are done! Thanks, @baks!!!

PR #387, opened 2015-06-19
CS0050 code fix

baks:
Fixes `CS0050` in #381 

I added `CS0050` to `InconsistentAccessibilityCodeFixProvider` with unit tests.
Also changed code action messages to use localizable strings.
Common methods used by `InconsistentAccessibilityInMethodParameter/(ReturnType)` were moved as extension methods.

### DIFF
--- a/src/CSharp/CodeCracker/Design/InconsistentAccessibility/InconsistentAccessibilityInMethodParameter.cs
+++ b/src/CSharp/CodeCracker/Design/InconsistentAccessibility/InconsistentAccessibilityInMethodParameter.cs
@@ -8,11 +8,14 @@ using System.Threading;
 using System.Threading.Tasks;
 using System.Text.RegularExpressions;
 using System.Globalization;
+using CodeCracker.Properties;
 
 namespace CodeCracker.CSharp.Design.InconsistentAccessibility
 {
     public sealed class InconsistentAccessibilityInMethodParameter : InconsistentAccessibilityInfoProvider
     {
+        private static readonly LocalizableString CodeActionMessage = new LocalizableResourceString(nameof(Resources.InconsistentAccessibilityInMethodParameter_CodeActionMessage), Resources.ResourceManager, typeof(Resources));
+
         public async Task<InconsistentAccessibilityInfo> GetInconsistentAccessibilityInfoAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
         {
             var result = new InconsistentAccessibilityInfo();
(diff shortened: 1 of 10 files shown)


src/CSharp/CodeCracker/Design/InconsistentAccessibility/InconsistentAccessibilityInMethodReturnType.cs
```diff
@@ -0,0 +1,34 @@
...
+            var result = new InconsistentAccessibilityInfo();
+            var syntaxRoot = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+            var methodThatRaisedError = syntaxRoot.FindNode(diagnostic.Location.SourceSpan).DescendantNodesAndSelf().OfType<MethodDeclarationSyntax>().FirstOrDefault();
+            if (methodThatRaisedError != null)
+            {
+                result.TypeToChangeAccessibility = methodThatRaisedError.ReturnType;
+                result.CodeActionMessage = string.Format(CodeActionMessage.ToString(), methodThatRaisedError.ReturnType, methodThatRaisedError.GetIdentifier().ValueText);
+                result.NewAccessibilityModifiers = methodThatRaisedError.CloneAccessibilityModifiers();
+            }
+
```
giggio: You will have to verify if the return type is in source, the same problem from `CC0051`.

MyDependence.dll:

````csharp
    public class DependedUpon
    {
        protected class A { }
    }
````

My code:

````csharp
    public class Class1 : ClassLibrary6.DependedUpon
    {
        public void Foo()
        {
            var a = new A();
        }
        public A Bar()
        {
            throw new Exception();
        }
    }
````

I can't change the accessibility of `DependedUpon.A` to be public, it is not in my source.

File: src/CSharp/CodeCracker/Design/InconsistentAccessibility/InconsistentAccessibilityInMethodParameter.cs [hunk 2]
```diff
@@ -8,11 +8,14 @@
 using System.Threading.Tasks;
 using System.Text.RegularExpressions;
 using System.Globalization;
+using CodeCracker.Properties;
 
 namespace CodeCracker.CSharp.Design.InconsistentAccessibility
 {
     public sealed class InconsistentAccessibilityInMethodParameter : InconsistentAccessibilityInfoProvider
     {
+        private static readonly LocalizableString CodeActionMessage = new LocalizableResourceString(nameof(Resources.InconsistentAccessibilityInMethodParameter_CodeActionMessage), Resources.ResourceManager, typeof(Resources));
```
giggio: The correct name for the resource would be `InconsistentAccessibilityInMethodParameter_Title`.

src/CSharp/CodeCracker/Design/InconsistentAccessibility/InconsistentAccessibilityInMethodReturnType.cs
baks: This is done by `IncosistentAccessibilityCodeFixProvider` in `RegisterCodeFixesAsync`

src/CSharp/CodeCracker/Design/InconsistentAccessibility/InconsistentAccessibilityInMethodParameter.cs
baks: I'll correct that

src/CSharp/CodeCracker/Design/InconsistentAccessibility/InconsistentAccessibilityInMethodReturnType.cs
giggio: Oh, right! Sorry about that.

Issue #368, opened 2015-06-01
Bug on CC0068 (RemovePrivateMethodNeverUsed) with Main


giggio:
`static void Main(string[] args)` will usually not have any outside reference. It should be an exception on RemovePrivateMethodNeverUsedAnalyzer.

According to [msdn](https://msdn.microsoft.com/en-us/library/acy3edy3.aspx?f=255&MSPPError=-2147217396):

> * The Main method is the entry point of an .exe program; it is where the program control starts and ends.
> * Main is declared inside a class or struct. Main must be static and it should not be public. (In the earlier example, it receives the default access of private.) The enclosing class or struct is not required to be static.
> * Main can either have a void or int return type.
> * The Main method can be declared with or without a string[] parameter that contains command-line arguments. When using Visual Studio to create Windows Forms applications, you can add the parameter manually or else use the Environment class to obtain the command-line arguments. Parameters are read as zero-indexed command-line arguments. Unlike C and C++, the name of the program is not treated as the first command-line argument.

What I have seen is that the accessibility of the method does not matter. We should exempt the method if and only if:

* It returns `void` or `int`.
* It is `static`.
* It is called exactly `Main` (not `main`, or `MAIN` or any other variation)
* it takes exactly on parameter of `string[]` or no parameter at all.

rtbortolin:
Can I do this one?

giggio:
Sure, @rtbortolin, go ahead.

PR #390, opened 2015-06-23
Fix bug on CC0068 (RemovePrivateMethodNeverUsed) with Main #368

rtbortolin:
Fixes CC0068 on #368

I've created and called a new method 'IsMainMethodEntryPoint' that verifies:
Method name is "Main";
Method is static;
Method returns 'Void' or 'Int32';
Method has more than 1 parameters;
Method has no parameter
Method has parameter type 'String[]'

And added respectives tests.


giggio:
Perfect! Tks @rtbortolin!

Issue #359, opened 2015-05-25
Bug: nameof analyzer (CC0021) being raised on self referencing initialization statement


giggio:
`CC0021` should not be raised within the statement that initializes a variable, like so:

````csharp
var str = SyntaxFactory.IdentifierName(nameof(str));
````

It also happens on properties, and so on:

````csharp
public int MyProperty { get; set; } = nameof(MyProperty);
````

PR #392, opened 2015-06-25
BugFix nameof analyzer (CC0021)

thorgeirk11:
Bug fix for #359. 
---
**Changes:**
- Added a check if the node is within a `VariableDeclaratorSyntax` or a `PropertyDeclarationSyntax`
- Added 2 new unit test to cover these cases.

I think this should fix the issue, maybe I am forgetting some edge case.


giggio:
@thorgeirk11 Nice!
Next time alert us that you are fixing the issue, so we don't have somebody else doing the same one.

thorgeirk11:
Will do :+1:

PR #394, opened 2015-06-25
Fix #393: make async method non async

### LINKED ISSUE #393: Make method non async (fix for CS1998)
The following method:
```
public static async Task<int> FooAsync()
{
    return 42;
}
```
Causes warning CS1998:

> This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread.

It can easily be fixed by replacing all `return x` statements with `Task.FromResult(x)` and removing the async keyword.

thomaslevesque:
if it lacks the await keyword.

There is only a codefix, no analyzer, since it's based on the compiler's
CS1998 warning. The codefix just removes the async keyword and replaces
all occurrences of `return x` with return `Task.FromResult(x)`.


giggio:
@thomaslevesque Nice one. I liked the rewriter approach, we have not used that before here.
I have added you to the collaborators team, so next time we can assign the issue to you.

PR #395, opened 2015-06-27
Removed parenthisis of empty constructor in object initializer code f…

### LINKED ISSUE #383: Using the refactoring for CC0008 will result in a warning CC0015
When using CC0008, the resulting object initializer will contain unnessecary parantheses and CC0015 will be trigered.

It would be better if parantheses where removed if the default constructor is used.

Example:

    var myFoo = new Foo();
    myFoo.bar = "foobar";

This will trigger CC0008.
After using it, the resulting code is

    var myFoo = new Foo()
    {
        bar = "foobar"
    };

this will trigger CC0015
After using it, the resulting code is

    var myFoo = new Foo
    {
        bar = "foobar"
    };

thorgeirk11:
Removed parenthesis of empty constructor in `ObjectInitializerCodeFixProvider`.
This prevents a CC0015 warning and fixes #383.

Sorry for not giving a heads up sooner. I was very quick to solve this issue.


giggio:
Good one.
Found a related bug, just reported it: #396. Maybe you want to tackle it too?

PR #401, opened 2015-07-02
Expanded CC0006 to deal with more array accessors

### LINKED ISSUE #318: Expand CC0006 (change for into a foreach)
Currently CC0006 is only raised on a specific pattern, with a for, and an assignment happening with the indexer like this:

````csharp
var array = new [] {1};
for (var i = 0; i < array.Length; i++)
{
    var item = array[i];
}
````

And it's code fix changes it to a foreach, like this:

````csharp
var array = new [] {1};
foreach (var item in array)
{
}
````

It is not currently able to deal with more array accessors, like this:

````csharp
var array = new [] {1};
var count = 0;
for (var i = 0; i < array.Length; i++)
{
    var item = array[i];
    count += array[i];
    Console.WriteLine(array[i]);
}
````

This could be changed to:

````csharp
var array = new [] {1};
var count = 0;
foreach (var item in array)
{
    count += item;
    Console.WriteLine(item);
}
````

This issue will expand both the analyzer and the code fix.

VB does not have this analyzer yet, so this would only be C#.

@albertomonteiro is working on it.

### PR DESCRIPTION
**Pull request by AlbertoMonteiro**
This PR Fix #318

Now CC0006 can deal with more array accessors, like this:
````csharp
var array = new [] {1};
var count = 0;
for (var i = 0; i < array.Length; i++)
{
    var item = array[i];
    count += array[i];
    Console.WriteLine(array[i]);
    count += array[array[i]];
}
````
CodeFix change old code to:
````csharp
var array = new [] {1};
var count = 0;
foreach (var item in array)
{
    count += item;
    Console.WriteLine(item);
    count += array[item];
}
````

### DIFF
--- a/src/CSharp/CodeCracker/Style/ForInArrayCodeFixProvider.cs
+++ b/src/CSharp/CodeCracker/Style/ForInArrayCodeFixProvider.cs
@@ -4,6 +4,8 @@ using Microsoft.CodeAnalysis.CodeFixes;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Formatting;
+using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Composition;
 using System.Linq;
@@ -37,7 +39,8 @@ namespace CodeCracker.CSharp.Style
             var condition = (BinaryExpressionSyntax)forStatement.Condition;
             var arrayAccessor = (MemberAccessExpressionSyntax)condition.Right;
             var arrayId = semanticModel.GetSymbolInfo(arrayAccessor.Expression).Symbol as ILocalSymbol;
-            var controlVarId = semanticModel.GetDeclaredSymbol(forStatement.Declaration.Variables.Single());
+            var forVariable = forStatement.Declaration.Variables.Single();
+            var controlVarId = semanticModel.GetDeclaredSymbol(forVariable);
             var arrayDeclarations = (from s in forBlock.Statements.OfType<LocalDeclarationStatementSyntax>()
                                      where s.Declaration.Variables.Count == 1
                                      let declaration = s.Declaration.Variables.First()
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/CSharp/CodeCracker/Style/ForInArrayCodeFixProvider.cs [hunk 1]
```diff
@@ -50,10 +53,27 @@ where controlVarId.Equals(initSymbol)
...
                 .WithLeadingTrivia(forStatement.GetLeadingTrivia())
                 .WithTrailingTrivia(forStatement.GetTrailingTrivia())
                 .WithAdditionalAnnotations(Formatter.Annotation);
+
+            var foreachVariable = SyntaxFactory.IdentifierName(forEachStatement.Identifier.Text);
+            Func<ForEachStatementSyntax, List<ElementAccessExpressionSyntax>> getElementsAccessorsToReplace;
+            getElementsAccessorsToReplace = (foreachStatement) => forEachStatement.DescendantNodes()
+                            .OfType<ElementAccessExpressionSyntax>()
+                            .Where(eae => eae.ArgumentList.Arguments.Any(a => a.Expression.ToFullString().Trim() == forVariable.Identifier.ToFullString().Trim()))
+                            .ToList();
```
giggio: @AlbertoMonteiro This will lead to a bug when there is an accessor that is being applied to a different array. For example:

````csharp
for (var i = 0; i < array.Length; i++)
{
    var item = array[i];
    count += array[i];
    anotherCount += anotherArray[i];
}
````

This will incorrectly be changed to:

````csharp
foreach (var item in array)
{
    count += item;
    anotherCount += item;
}
````

giggio: Or maybe not. I will run it. The analyzer might not raise a diagnostic for this condition.

AlbertoMonteiro: @giggio I will add a test to ensure that doesnt happen.

AlbertoMonteiro: @giggio You right, code fix changed incorrectly, I will fix it now

AlbertoMonteiro: @giggio Fixed,

giggio: Yes, there was a problem. I am doing a quick fix and I will merge.

AlbertoMonteiro: I did some refactor to improve code

giggio: @AlbertoMonteiro I found that that last was not correct, as it would introduce a compilation error. I changed the analyzer to disallow accessors that are not from the same symbol. Other types of use of the for initializer variable were already forbidden. This bug was there already before you worked your fix.

giggio:
@AlbertoMonteiro Tks for the improvement! :)

src/CSharp/CodeCracker/Style/ForInArrayCodeFixProvider.cs
AlbertoMonteiro: Yeah, didn't saw that. Thanks.

PR #402, opened 2015-07-02
Ignore Disposable objects created on return statements

### LINKED ISSUE #399: Ignore Disposables created on return statements
When an ```IDisposable``` object is created on the return statement of a method, the ```DisposableVariableNotDisposedAnalyzer``` creates a ```Diagnostic``` saying that the said object should be disposed. Like in the snippet below:

```csharp
public class DisposableClass : System.IDisposable  { }
public class ActualClass
{
    public DisposableClass Method()
    {
        return new DisposableClass(); 
     }
}
```

Trying to Fix Occurences of said Diagnostic causes VS to crash when method ```CreateUsing``` (class ```DisposableVariableNotDisposedCodeFixProvider```) is called. The crash is caused by an ```InvalidCastException``` thrown by ```CreateUsing``` here:

```csharp
newRoot = CreateRootWithUsing(root, (ExpressionStatementSyntax)objectCreation.Parent, u => u.WithExpression(objectCreation));
```
Since the disposable object is being created and returned, the ```Parent``` property of ```objectCreation``` in the code above is actually of type ```ReturnStatementSyntax```, causing the ```InvalidCastException``` to be thrown.

[I have already fixed this in my fork](https://github.com/ericcastoldi/code-cracker/commit/214cadb04659e23bd5f74e863b3d9220ee2a21d2), so the ```DisposableVariableNotDisposedAnalyzer``` ignores nodes
whose Parents are of type ```ReturnStatementSyntax```. But i'm not sure if i should submit the pull request yet because i guess people would have to discuss about it first, and maybe someone has a better approach.

I'll wait for your comments before submitting the pull request.

ericcastoldi:
Fixed the ```DisposableVariableNotDisposedAnalyzer``` so it ignores nodes
whose Parents are of type ```ReturnStatementSyntax```

Closes #399.

### DIFF
--- a/src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedAnalyzer.cs
@@ -37,6 +37,7 @@ namespace CodeCracker.CSharp.Usage
             var objectCreation = context.Node as ObjectCreationExpressionSyntax;
             if (objectCreation == null) return;
             if (objectCreation?.Parent is UsingStatementSyntax) return;
+            if (objectCreation?.Parent is ReturnStatementSyntax) return;
 
             var semanticModel = context.SemanticModel;
             var type = semanticModel.GetSymbolInfo(objectCreation.Type).Symbol as INamedTypeSymbol;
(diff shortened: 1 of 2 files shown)


ericcastoldi:
@giggio  AppVeyor's build failed because some server called from the build script is unavailable. Is there a way to force run another build?

AlbertoMonteiro:
@ericcastoldi yeah, you can ammend some change in your commit, and do a force push.
Add some NewLine, commit with ammend, then remove line and do another commit with ammend

ericcastoldi:
Oh, of course, i should have thought of that... thanks @AlbertoMonteiro!

AlbertoMonteiro:
@ericcastoldi you are welcome :yum:

carloscds:
@ericcastoldi Good job!

ericcastoldi:
Thanks @carloscds :smiley:

PR #403, opened 2015-07-06
Implemented analyzer to fix xml docs parameters

### LINKED ISSUE #357: Verify if xml docs have the correct parameters
As fixed on 526aa919a3afa369890ee9adfd81aae2216c5921.

Should work on methods and constructors.
Diagnostic should only be raised if there is already some xml documentation present.

### Removing unexistent param:

Before (`spans` does not exist):

````csharp
         /// </summary>
         /// <param name="analyzer">The analyzer to run on the documents</param>
         /// <param name="documents">The Documents that the analyzer will be run on</param>
         /// <param name="spans">Optional TextSpan indicating where a Diagnostic will be found</param>
         /// <returns>An IEnumerable of Diagnostics that surfaced in teh source code, sorted by Location</returns>
         protected async static Task<Diagnostic[]> GetSortedDiagnosticsFromDocumentsAsync(DiagnosticAnalyzer analyzer, Document[] documents)
         {
````

After:

````csharp
         /// </summary>
         /// <param name="analyzer">The analyzer to run on the documents</param>
         /// <param name="documents">The Documents that the analyzer will be run on</param>
         /// <returns>An IEnumerable of Diagnostics that surfaced in teh source code, sorted by Location</returns>
         protected async static Task<Diagnostic[]> GetSortedDiagnosticsFromDocumentsAsync(DiagnosticAnalyzer analyzer, Document[] documents)
         {
````

### Creating a missing parameter:

Before (note that `workspace` does not appear on the xml docs):

````csharp
         /// </summary>
         /// <param name="sources">Classes in the form of strings</param>
         /// <param name="language">The language the source code is in</param>
         /// <returns>A Project created out of the Douments created from the source strings</returns>
         public static Project CreateProject(string[] sources,
             out AdhocWorkspace workspace, string language = LanguageNames.CSharp)
````

After (with `workspace`):

````csharp
         /// </summary>
         /// <param name="sources">Classes in the form of strings</param>
         /// <param name="language">The language the source code is in</param>
         /// <param name="workspace">todo: describe workspace parameter on CreateProject</param>
         /// <returns>A Project created out of the Douments created from the source strings</returns>
         public static Project CreateProject(string[] sources,
             out AdhocWorkspace workspace, string language = LanguageNames.CSharp)
````

When including a parameter leave the description in the format:

````
todo: describe {parameter} parameter on {method}
````

The code actions listed should be:

 * One for each missing parameter
 * One for all parameters that are missing

Severity: `Warning`
Diagnostic id: `CC0090`
Category: `Maintenability`

### PR DESCRIPTION
**Pull request by AlbertoMonteiro**
I created analyzer and 2 code fixes, one for missing parameters and another to unexisting parameters.

Tests are ok, but I am facing a problem when testing running VS 

But when I press <kbd>CTRL</kbd>+<kbd>.</kbd> I get a error in VS.
**"One or more erros ocurred."**

When I checked to see all exceptions in VS I got StackTrace and got this exception:
````
System.InvalidCastException occurred
Message: Exception thrown: 'System.InvalidCastException' in mscorlib.dll
Additional information: Unable to cast object of type 'System.Reflection.RuntimeMethodInfo' to type 'System.Reflection.ConstructorInfo'.
````

I tried a lot things, but couldnt fix, I need help.
Fix #357

### DIFF
--- a/src/CSharp/CodeCracker/CodeCracker.csproj
+++ b/src/CSharp/CodeCracker/CodeCracker.csproj
@@ -71,6 +71,9 @@
     <Compile Include="Refactoring\IntroduceFieldFromConstructorCodeFixProvider.cs" />
     <Compile Include="Refactoring\ParameterRefactoryAnalyzer.cs" />
     <Compile Include="Refactoring\ParameterRefactoryCodeFixProvider.cs" />
+    <Compile Include="Maintainability\XmlDocumentationAnalyzer.cs" />
+    <Compile Include="Maintainability\XmlDocumentationCreateMissingParametersCodeFixProvider.cs" />
+    <Compile Include="Maintainability\XmlDocumentationRemoveNonExistentParametersCodeFixProvider.cs" />
     <Compile Include="Reliability\UseConfigureAwaitFalseAnalyzer.cs" />
     <Compile Include="Reliability\UseConfigureAwaitFalseCodeFixProvider.cs" />
     <Compile Include="Style\ConsoleWriteLineAnalyzer.cs" />
@@ -99,6 +102,7 @@
     <Compile Include="Style\RemoveTrailingWhitespaceCodeFixProvider.cs" />
     <Compile Include="Style\StringFormatCodeFixProvider.cs" />
     <Compile Include="Refactoring\NumericLiteralAnalyzer.cs" />
+    <Compile Include="Maintainability\XmlDocumentationCodeFixProvider.cs" />
     <Compile Include="Usage\AbstractClassShouldNotHavePublicCtorsCodeFixProvider.cs" />
     <Compile Include="Usage\AbstractClassShouldNotHavePublicCtorsAnalyzer.cs" />
     <Compile Include="Usage\ArgumentExceptionAnalyzer.cs" />
@@ -250,7 +254,6 @@
   <ItemGroup>
     <Folder Include="Globalization\" />
     <Folder Include="Interoperability\" />
-    <Folder Include="Maintainability\" />
     <Folder Include="Mobility\" />
     <Folder Include="Naming\" />
     <Folder Include="Portability\" />
--- a/src/CSharp/CodeCracker/Maintainability/XmlDocumentationAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Maintainability/XmlDocumentationAnalyzer.cs
@@ -1,0 +1,63 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using CodeCracker.Properties;
+
+namespace CodeCracker.CSharp.Maintainability
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class XmlDocumentationAnalyzer : DiagnosticAnalyzer
+    {
+        internal static readonly LocalizableString Title = new LocalizableResourceString(nameof(Resources.XmlDocumentationAnalyzer_Title), Resources.ResourceManager, typeof(Resources));
+
+        internal static DiagnosticDescriptor Rule = new DiagnosticDescriptor(
+            DiagnosticId.XmlDocumentation.ToDiagnosticId(),
+            Title,
+            Title,
+            SupportedCategories.Maintainability,
+            DiagnosticSeverity.Info,
+            isEnabledByDefault: true,
+            helpLinkUri: HelpLink.ForDiagnostic(DiagnosticId.XmlDocumentation));
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context) => context.RegisterSyntaxNodeAction(Analyzer, SyntaxKind.SingleLineDocumentationCommentTrivia);
+
+        private static void Analyzer(SyntaxNodeAnalysisContext context)
+        {
+            var documentationNode = (DocumentationCommentTriviaSyntax)context.Node;
+            var method = documentationNode.AncestorsAndSelf().OfType<MethodDeclarationSyntax>().First();
+            var methodParameters = method.ParameterList.Parameters;
+            var xElementsWitAttrs = documentationNode.Content.OfType<XmlElementSyntax>()
+                                    .Where(xEle => xEle.StartTag.Name.LocalName.ValueText == "param")
+                                    .SelectMany(xEle => xEle.StartTag.Attributes, (xEle, attr) => (XmlNameAttributeSyntax)attr);
+
+            var keys = methodParameters.Select(parameter => parameter.Identifier.ValueText)
+                .Union(xElementsWitAttrs.Select(x => x.Identifier.Identifier.ValueText))
+                .ToImmutableHashSet();
+
+            var paramterWithDocParameter = (from key in keys
+                                            let Parameter = methodParameters.FirstOrDefault(p => p.Identifier.ValueText == key)
+                                            let DocParameter = xElementsWitAttrs.FirstOrDefault(p => p.Identifier.Identifier.ValueText == key)
+                                            select new { Parameter, DocParameter });
+
+            if (paramterWithDocParameter.Any(p => p.Parameter == null))
+            {
+                var properties = new Dictionary<string, string> {["kind"] = "nonexistentParam" }.ToImmutableDictionary();
+                var diagnostic = Diagnostic.Create(Rule, documentationNode.GetLocation(), properties);
+                context.ReportDiagnostic(diagnostic);
+            }
+
+            if (paramterWithDocParameter.Any(p => p.DocParameter == null))
+            {
+                var properties = new Dictionary<string, string> { ["kind"] = "missingDoc" }.ToImmutableDictionary();
+                var diagnostic = Diagnostic.Create(Rule, documentationNode.GetLocation(), properties);
+                context.ReportDiagnostic(diagnostic);
+            }
+        }
+    }
+}
--- a/src/CSharp/CodeCracker/Maintainability/XmlDocumentationCodeFixProvider.cs
+++ b/src/CSharp/CodeCracker/Maintainability/XmlDocumentationCodeFixProvider.cs
@@ -1,0 +1,47 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace CodeCracker.CSharp.Maintainability
+{
+    public abstract class XmlDocumentationCodeFixProvider : CodeFixProvider
+    {
+        public sealed override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(DiagnosticId.XmlDocumentation.ToDiagnosticId());
+
+        public abstract SyntaxNode FixParameters(MethodDeclarationSyntax method, SyntaxNode root);
+
+        protected async Task<Document> FixParametersAsync(Document document, Diagnostic diagnostic, CancellationToken c)
+        {
+            var root = await document.GetSyntaxRootAsync(c).ConfigureAwait(false);
+            var documentationNode = root.FindToken(diagnostic.Location.SourceSpan.Start).Parent.AncestorsAndSelf().OfType<MethodDeclarationSyntax>().First();
+            var newRoot = FixParameters(documentationNode, root);
+            var newDocument = document.WithSyntaxRoot(newRoot);
+
+            return newDocument;
+        }
+
+        protected static IEnumerable<Tuple<ParameterSyntax, Tuple<XmlElementSyntax, XmlNameAttributeSyntax>>> GetMethodParametersWithDocParameters(MethodDeclarationSyntax method, DocumentationCommentTriviaSyntax documentationNode)
+        {
+            var methodParameters = method.ParameterList.Parameters;
+
+            var xElementsWitAttrs = documentationNode.Content.OfType<XmlElementSyntax>()
+                                    .Where(xEle => xEle.StartTag.Name.LocalName.ValueText == "param")
+                                    .SelectMany(xEle => xEle.StartTag.Attributes, (xEle, attr) => new Tuple<XmlElementSyntax, XmlNameAttributeSyntax>(xEle, (XmlNameAttributeSyntax)attr));
+
+            var keys = methodParameters.Select(parameter => parameter.Identifier.ValueText)
+                .Union(xElementsWitAttrs.Select(x => x.Item2.Identifier.Identifier.ValueText))
+                .ToImmutableHashSet();
+
+            return (from key in keys
+                    let Parameter = methodParameters.FirstOrDefault(p => p.Identifier.ValueText == key)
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

File: src/CSharp/CodeCracker/Refactoring/XmlDocumentationAnalyzer.cs [hunk 1]
```diff
@@ -0,0 +1,65 @@
...
+        internal const string Category = SupportedCategories.Refactoring;
+
+        internal static DiagnosticDescriptor XmlDocumentationRule = new DiagnosticDescriptor(
+            DiagnosticId.XmlDocumentation.ToDiagnosticId(),
+            Title,
+            MessageFormat,
+            SupportedCategories.Refactoring,
+            DiagnosticSeverity.Warning,
+            isEnabledByDefault: true,
+            helpLinkUri: HelpLink.ForDiagnostic(DiagnosticId.XmlDocumentation));
```
giggio: Category should be maintenability. Needs to change the `DiagnosticSeverity` and file locations.

src/CSharp/CodeCracker/Refactoring/XmlDocumentationCreateMissingParametersCodeFixProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        private static SyntaxList<XmlAttributeSyntax> CreateXmlAttributes(string paramenterName)
+        {
+            var syntaxToken = SyntaxFactory.Token(DoubleQuoteToken);
+            var xmlAttributeName = SyntaxFactory.XmlName(SyntaxFactory.Identifier(@"name"));
+
+            return SyntaxFactory.SingletonList<XmlAttributeSyntax>(SyntaxFactory.XmlNameAttribute(xmlAttributeName, syntaxToken, SyntaxFactory.IdentifierName(paramenterName), syntaxToken));
+        }
+
+        public override Task RegisterCodeFixesAsync(CodeFixContext context)
```
giggio: `RegisterCodeFixesAsync` should be sealed.

src/CSharp/CodeCracker/Refactoring/XmlDocumentationRemoveUnexistentParametersCodeFixProvider.cs
```diff
@@ -0,0 +1,48 @@
...
+            var allNodesToRemove = nodesToRemove.Cast<SyntaxNode>().Union(xmlTextNodesToRemove);
+            return allNodesToRemove;
+        }
+
+        public override Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var diagnostic = context.Diagnostics.First();
+            if (diagnostic.Properties["kind"] == "unexistent")
+                context.RegisterCodeFix(CodeAction.Create(MessageFormat, c => FixParametersAsync(context.Document, diagnostic, c)), diagnostic);
+            return Task.FromResult(0);
```
giggio: `RegisterCodeFixesAsync` should be sealed.

src/CSharp/CodeCracker/Refactoring/XmlDocumentationCodeFixProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace CodeCracker.CSharp.Refactoring
+{
+    public abstract class XmlDocumentationCodeFixProvider : CodeFixProvider
+    {
+        public sealed override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(DiagnosticId.XmlDocumentation.ToDiagnosticId());
+
+        public abstract SyntaxNode FixParameters(MethodDeclarationSyntax method, SyntaxNode root);
+
+        public abstract string MessageFormat { get; }
```
giggio: No need for this property to exist on the base class, it is not used here. It can be a const string field (not a property) on the derived classes

src/CSharp/CodeCracker/Refactoring/XmlDocumentationCreateMissingParametersCodeFixProvider.cs
sharwell: :question: Why?

src/CSharp/CodeCracker/Refactoring/XmlDocumentationCodeFixProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+            var methodParameters = method.ParameterList.Parameters;
+
+            var xElementsWitAttrs = documentationNode.Content.OfType<XmlElementSyntax>()
+                                    .Where(xEle => xEle.StartTag.Name.LocalName.ValueText == "param")
+                                    .SelectMany(xEle => xEle.StartTag.Attributes, (xEle, attr) => new Tuple<XmlElementSyntax, XmlNameAttributeSyntax>(xEle, (XmlNameAttributeSyntax)attr));
+
+            var methodParameterLookUp = methodParameters.ToLookup(parameter => parameter.Identifier.ValueText);
+            var docParameterLookUp = xElementsWitAttrs.ToLookup(x => x.Item2.Identifier.Identifier.ValueText);
+
+            var keys = methodParameterLookUp.Select(p => p.Key).Union(docParameterLookUp.Select(p => p.Key)).ToImmutableHashSet();
```
giggio: It does not seem necessary to create 2 lookups if the only reason for their existence is to later create a union on their keys. Analyzers are all about performance, simply create a hash from the parameter identifiers and xelements without creating the intermediate lookup.

src/CSharp/CodeCracker/Refactoring/XmlDocumentationAnalyzer.cs
```diff
@@ -0,0 +1,65 @@
...
+            var method = documentationNode.AncestorsAndSelf().OfType<MethodDeclarationSyntax>().First();
+            var methodParameters = method.ParameterList.Parameters;
+            var xElementsWitAttrs = documentationNode.Content.OfType<XmlElementSyntax>()
+                                    .Where(xEle => xEle.StartTag.Name.LocalName.ValueText == "param")
+                                    .SelectMany(xEle => xEle.StartTag.Attributes, (xEle, attr) => (XmlNameAttributeSyntax)attr);
+
+            var methodParameterLookUp = methodParameters.ToLookup(parameter => parameter.Identifier.ValueText);
+            var docParameterLookUp = xElementsWitAttrs.ToLookup(attr => attr.Identifier.Identifier.ValueText);
+
+            var keys = methodParameterLookUp.Select(p => p.Key).Union(docParameterLookUp.Select(p => p.Key)).ToImmutableHashSet();
```
giggio: It does not seem necessary to create 2 lookups if the only reason for their existence is to later create a union on their keys. Analyzers are all about performance, simply create a hash from the parameter identifiers and xelements without creating the intermediate lookup.

src/CSharp/CodeCracker/Refactoring/XmlDocumentationAnalyzer.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                var properties = new Dictionary<string, string> {["kind"] = "unexistent" }.ToImmutableDictionary();
+                var diagnostic = Diagnostic.Create(XmlDocumentationRule, documentationNode.GetLocation(), properties, properties["kind"]);
+                context.ReportDiagnostic(diagnostic);
+            }
+
+            if (paramterWithDocParameter.Any(p => p.DocParameter == null))
+            {
+                var properties = new Dictionary<string, string> { ["kind"] = "missing" }.ToImmutableDictionary();
+                var diagnostic = Diagnostic.Create(XmlDocumentationRule, documentationNode.GetLocation(), properties, properties["kind"]);
```
giggio: No need for the last parameter on the `Diagnostic.Create` invocation, you are not using it on the `MessageFormat`.

giggio: I am finding it difficult to follow the `missing` and `unexistent` texts. They almost mean the same thing. Can you try to be more descriptive? Maybe `missingDoc` and `nonexistentParam`, or something like that?

src/CSharp/CodeCracker/Refactoring/XmlDocumentationRemoveUnexistentParametersCodeFixProvider.cs
```diff
@@ -0,0 +1,48 @@
...
+using System;
+using System.Collections.Generic;
+using System.Composition;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace CodeCracker.CSharp.Refactoring
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp, nameof(XmlDocumentationCodeFixProvider)), Shared]
+    public class XmlDocumentationRemoveUnexistentParametersCodeFixProvider : XmlDocumentationCodeFixProvider
```
giggio: Use `NonExistent` instead of `Unexistent`.

src/CSharp/CodeCracker/Refactoring/XmlDocumentationCodeFixProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+        protected static IEnumerable<Tuple<ParameterSyntax, Tuple<XmlElementSyntax, XmlNameAttributeSyntax>>> GetMethodParametersWithDocParameters(MethodDeclarationSyntax method, DocumentationCommentTriviaSyntax documentationNode)
+        {
+            var methodParameters = method.ParameterList.Parameters;
+
+            var xElementsWitAttrs = documentationNode.Content.OfType<XmlElementSyntax>()
+                                    .Where(xEle => xEle.StartTag.Name.LocalName.ValueText == "param")
+                                    .SelectMany(xEle => xEle.StartTag.Attributes, (xEle, attr) => new Tuple<XmlElementSyntax, XmlNameAttributeSyntax>(xEle, (XmlNameAttributeSyntax)attr));
+
+            var methodParameterLookUp = methodParameters.ToLookup(parameter => parameter.Identifier.ValueText);
+            var docParameterLookUp = xElementsWitAttrs.ToLookup(x => x.Item2.Identifier.Identifier.ValueText);
```
giggio: You don't need the Tuple if all you are doing is accessing `Item2`. Simply return it from the `SelectMany` above.

src/CSharp/CodeCracker/Refactoring/XmlDocumentationCodeFixProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+        public sealed override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(DiagnosticId.XmlDocumentation.ToDiagnosticId());
+
+        public abstract SyntaxNode FixParameters(MethodDeclarationSyntax method, SyntaxNode root);
+
+        public abstract string MessageFormat { get; }
+
+        protected async Task<Document> FixParametersAsync(Document document, Diagnostic diagnostic, CancellationToken c)
+        {
+            var root = await document.GetSyntaxRootAsync(c).ConfigureAwait(false);
+            var documentationNode = (MethodDeclarationSyntax)root.FindNode(diagnostic.Location.SourceSpan);
```
giggio: I caught an exception on this line:

````cshap
var documentationNode = (MethodDeclarationSyntax)root.FindNode(diagnostic.Location.SourceSpan);
````

Details:

````
Exception thrown: 'System.InvalidCastException' in CodeCracker.CSharp.dll

Additional information: Unable to cast object of type 'Microsoft.CodeAnalysis.CSharp.Syntax.PredefinedTypeSyntax' to type 'Microsoft.CodeAnalysis.CSharp.Syntax.MethodDeclarationSyntax'.
````

The problem is the node being found is the return type of the method. Even though the return type has a method **parent**, it is not the `MethodDeclaration` that is being found by the `FindNode` method. The most specific type is always found. Try to use `root.FindToken(diagnostic.Location.SourceSpan.Start).Parent.AncestorsAndSelf().OfType<MethodDeclarationSyntax>().First()` instead.

src/CSharp/CodeCracker/Refactoring/XmlDocumentationCreateMissingParametersCodeFixProvider.cs
giggio: @sharwell 3 reasons: style, recommendation and performance.
Style: All our `RegisterCodeFixesAsync` are sealed.
Recommendation: this is how Microsoft is recommending on their own template, and this is how it is done on Roslyn itself.
Performance: while the compiler usually emits IL with a `callvirt` to sealed and non-virtual method calls, the method is still marked `final`, which is a hint for the JIT later for possible optimizations, like method inlining.

src/CSharp/CodeCracker/Refactoring/XmlDocumentationAnalyzer.cs
AlbertoMonteiro: Which  ````DiagnosticSeverity````  should I set?

src/CSharp/CodeCracker/Refactoring/XmlDocumentationCodeFixProvider.cs
AlbertoMonteiro: I do it in method GetAllNodesToRemove on ````XmlDocumentationRemoveNonExistentParametersCodeFixProvider````

````csharp
var nodesToRemove = paramterWithDocParameter.Where(p => p.Item1 == null).Select(x => x.Item2.Item1).ToList();
````

AlbertoMonteiro:
@giggio I made many commits to fix your observations. When everything is aproved I will squash them all.

AlbertoMonteiro:
@giggio How did you catch this exception in comment: https://github.com/code-cracker/code-cracker/pull/403#discussion_r33905132 ?
After these commits to fix your observations, I ran VS and it fixed the <kbd>CTRL</kbd> + <kbd>.</kbd>

I walk through all my new commits and discovered that <kbd>CTRL</kbd> + <kbd>.</kbd> bug stoped in https://github.com/AlbertoMonteiro/code-cracker/commit/ebdae16c005a967b782bd921c514466a21072721

So I started to mix the changes that I did in this commit, and I discovered that if I use the namespace **CodeCracker.CSharp.Refactoring** it will break when running VS.
So I moved to parent commit(https://github.com/AlbertoMonteiro/code-cracker/commit/150a201a5a21f39b32185b63e049261f3a8b1357) and ran VS again, the bug happend, stoped, and change the namespace from **Refactoring** to **AnyElse** and the problem was gone.

Why the simple namespace name change, breack when running on VS?

giggio:
@AlbertoMonteiro You set the exception settings to catch all .NET Exceptions. If you don't do that, VS will not stop to let you inspect the exception. I don't know why, it should stop.
BTW, you also need to update the analyzer and code fix to use Resources for globalization support. See https://github.com/code-cracker/code-cracker/blob/cc44ad2419e25d794471b83867fbd08877af25dd/src/CSharp/CodeCracker/Style/ConsoleWriteLineAnalyzer.cs and the corresponding code fix.

giggio:
@AlbertoMonteiro I can't see why the namespace would break the code. Try using the exception setting I talked about above and see if you can find what happened. Anyway, you are going to move it another namespace, so maybe this is moot.

src/CSharp/CodeCracker/Refactoring/XmlDocumentationCreateMissingParametersCodeFixProvider.cs
sharwell: :question: Why not just make the class sealed?

AlbertoMonteiro:
@giggio commited the globalization support.
About the namespace breaking, I will record a vídeo and upload on youtube , so you can see this weird behavior.

src/CSharp/CodeCracker/Refactoring/XmlDocumentationCodeFixProvider.cs
giggio: Ok, got it.

giggio:
@AlbertoMonteiro Good job!
I did some small updates, check e1d2286 for more info. I got some edge cases that needed taking care of and the test file was on the wrong directory.

AlbertoMonteiro:
@giggio Thanks.
I will pay more antention next time!

Issue #410, opened 2015-07-21
Incorrect spacing in string interpolation (CC0048)


thorgeirk11:
Incorrect spacing in string interpolation (CC0048)
---
The code fix for  [CC0048](https://code-cracker.github.io/diagnostics/CC0048.html) transforms the following code 
```CSharp
var myString = string.Format(@"{0}, {1}, {2}, {3}", 0, 1,
                             2,
                             3);
```
Into 
```CSHarp
var myString = $@"{0}, {1}, {                             2}, {                             3}";
```
Since spaces inside an expression have no effect the desired result should be
```CSHarp
var myString = $@"{0}, {1}, {2}, {3}";
```

PR #414, opened 2015-07-22
Fix issue 410

carloscds:
Issue #410


thorgeirk11:
Nice job. Could you also add a test for the following scenario

 ```CSharp
var myString = string.Format("{0}, {1}, {2}, {3}", 0, /* Just a simple Comment */1,
                            /* Important Comment */ 2,
                            3 /* Another important Comment */);
```
Just to be sure the comments don't get removed.

giggio:
@thorgeirk11 I added it and it worked. I will commit it.

PR #421, opened 2015-07-23
Fix #65 (Change from as operator to direct cast or the opposite)

### LINKED ISSUE #65: Change from as operator to direct cast or the opposite
This:

````csharp
var nullLiteral = returnElse.Expression as LiteralExpressionSyntax;
````

Becomes:

````csharp
var nullLiteral = (LiteralExpressionSyntax)returnElse.Expression;
````

Or the opposite, turn a direct cast into an expression with the `as` operator.

Severity is `Hidden`.
Diagnostic id is `CC0028`.
Category: `Refactoring`

### PR DESCRIPTION
**Pull request by thomaslevesque**

### CODE REVIEW

**Comment by thomaslevesque** [comment]
Looks like Coveralls.io is having trouble, the job backlog is up to 30,000... might be a while before it processes this PR

**Comment by giggio** [comment]
@thomaslevesque Please send the PR to the `vnext` branch. `Master` is for stabilization only at the moment while we prepare for 1.0. They are both at the same commit, so it is only a matter of updating the PR.

**Comment by giggio** [comment]
I have comments as well, I will comment directly on the source.

File: src/CSharp/CodeCracker/Refactoring/ChangeAsToCastAnalyzer.cs [hunk 1]
```diff
@@ -0,0 +1,51 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace CodeCracker.CSharp.Refactoring
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class ChangeAsToCastAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string Title = "Toggle between safe cast ('as') and direct cast.";
+        internal const string MessageFormat = "Toggle between safe cast ('as') and direct cast.";
```
giggio: Use resources for these strings. See the `NameOfAnalyzer` with the use of `LocalizableResourceString`.

src/CSharp/CodeCracker/Refactoring/ChangeAsToCastCodeFixProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var diagnostic = context.Diagnostics.First();
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var node = root.FindNode(diagnostic.Location.SourceSpan);
+            if (node.Kind() == SyntaxKind.AsExpression)
+                context.RegisterCodeFix(CodeAction.Create("Change safe cast ('as') to direct cast", ct => ChangeAsToCastAsync(context.Document, root, node)), diagnostic);
+            else if (node.Kind() == SyntaxKind.CastExpression)
+                context.RegisterCodeFix(CodeAction.Create("Change direct cast to safe cast ('as')", ct => ChangeCastToAsAsync(context.Document, root, node)), diagnostic);
```
giggio: You have to inform the `equivalenceKey`. See the other analizers when calling `CodeAction.Create` for an example.
Also, use Resources here as well, see the `NameOfCodeFixProvider` for an example.

giggio: Instead of doing work on the `RegisterCodeFixesAsync` use properties to pass info from the analyzer to the code fix. See the `RemoveWhereWhenItIsPossibleAnalyzer` (lines 58 and 59) and `RemoveWhereWhenItIsPossibleCodeFixProvider` (line 25).
The idea is that `RegisterCodeFixesAsync` runs as fast as possible.

thomaslevesque:
> @thomaslevesque Please send the PR to the vnext branch. Master is for stabilization only at the moment while we prepare for 1.0. They are both at the same commit, so it is only a matter of updating the PR.

OK, but then I have to re-create the PR, right?

src/CSharp/CodeCracker/Refactoring/ChangeAsToCastCodeFixProvider.cs
thomaslevesque: > You have to inform the equivalenceKey. See the other analizers when calling CodeAction.Create for an example.

Looks like you commited this change after I started making my changes... I'll rebase my branch.

thomaslevesque:
OK, all done

I didn't know how to pass data between the analyzer and the code fix, thanks for letting me know ;)

I'm closing this PR and creating another, because it doesn't seem possible to change the target branch.

Issue #418, opened 2015-07-23
CC064 invalid error


SJMakin:
I believe that an error is being generated incorrectly

The code:

    Public Shared Function ToEnum(Of TEnum)(ByVal value As Object) As TEnum
        Dim rawValue As String = value.ToString()
        Dim result As TEnum
        Try
            result = DirectCast([Enum].Parse(GetType(TEnum), rawValue), TEnum)
        Catch ex As ArgumentException
            result = Nothing
        End Try

        Return result
    End Function

The issue is that 'GetType(TEnum)' is being shown as CC0064 Object reference not set to instance of an object, when it is valid.

@adraut is working on it.

giggio:
Bug confirmed. `CC0064` is incorrectly being raised on `Enum.Parse`. A simpler repro:

````vb
    Public Sub Foo()
        [Enum].Parse(GetType(String), "")
    End Sub
````

adraut:
@giggio Should I add to the analyzer to also cover the TryParse method or should that be done in a different issue?

giggio:
@adraut Seems like a different issue to me.

mc0re:
Same goes for other Parse() functions (Integer.Parse, Double.Parse, Byte.Parse).

giggio:
@mc0re This should not affect other parse functions after the fix on #418. We have not yet published this update. I am thinking of releasing an RC2 before 1.0, so this bugs already fixed don't have to be waiting.

PR #427, opened 2015-07-26
Fix Bug #418

adraut:
Fix Bug #418


giggio:
@adraut This does not seem correct. The correction should be on the `MethodChecker` class. If you look into the `AnalyzeMethod` for the C# version you will see that it calls `MethodFullNameNotFound`, and the VB one does not. I guess the problem is there. There should be no change at `IPAddressAnalyzer`.
Please verify that so we can move forward.
The tests are fine, but I guess that the C# one was already passing, as the implementation in C# seems correct.

adraut:
@giggio You were correct.  Closing and will submit a corrected PR.

PR #435, opened 2015-07-27
Add ValidateColor Analizer

### LINKED ISSUE #1: Validate color from System.Drawing.ColorTranslator.FromHtml
This:

````csharp
System.Drawing.ColorTranslator.FromHtml("wrong color");
````

Will be marked as an `Error`. Category is `Usage`.
Diagnostic id is `CC0053`

How to implement it is being discussed below, we still have to find an optimal way to do it, as we cannot call `ColorTranslator` from the Code Cracker project and see if it throws, as CC is a PCL, and `System.Drawing` is not available.

@dgmachado is working on it.

### PR DESCRIPTION
**Pull request by dgmachado**
Closes #1

Validate color from System.Drawing.ColorTranslator.FromHtml

Changed the target branch to vnext

### CODE REVIEW

File: src/CSharp/CodeCracker/Usage/ValidateColorAnalyzer.cs [hunk 1]
```diff
@@ -0,0 +1,2281 @@
...
+using System.Globalization;
+
+namespace CodeCracker.CSharp.Usage
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class ValidateColorAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string Category = SupportedCategories.Usage;
+        internal const string Message = "Your htmlColor value doesn't exist.";
+        internal const string Title = "Color validation";
```
giggio: Needs to use `LocalizableString`. See `NameofAnalyzer` as an example.

src/CSharp/CodeCracker/Usage/ValidateColorAnalyzer.cs
```diff
@@ -0,0 +1,2281 @@
...
+
+        private static void Analyzer(SyntaxNodeAnalysisContext context)
+        {
+            if (context.IsGenerated()) return;
+            var invocationExpression = (InvocationExpressionSyntax)context.Node;
+
+            var memberExpresion = invocationExpression.Expression as MemberAccessExpressionSyntax;
+            if (memberExpresion?.Name?.ToString() != "FromHtml") return;
+
+            if (memberExpresion.Expression as IdentifierNameSyntax != null)
```
giggio: Use `IsKind`, it is faster than checking types.

src/CSharp/CodeCracker/Usage/ValidateColorAnalyzer.cs
```diff
@@ -0,0 +1,2281 @@
...
+            {
+                if (((IdentifierNameSyntax)memberExpresion.Expression).Identifier.Text != nameof(ColorTranslator)) return;
+            }
+            else
+            {
+                if (((IdentifierNameSyntax)((MemberAccessExpressionSyntax)memberExpresion.Expression).Name).Identifier.Text != nameof(ColorTranslator)) return;
+            }
```
giggio: Instead of navigating the member access expression, it is better to as for the symbol, and be sure. There might be casts involved, etc, and the structure might not be exactly as you project in this `if else` case. See the `StringFormatAnalyzer`, it does exactly that.

src/CSharp/CodeCracker/Usage/ValidateColorAnalyzer.cs
```diff
@@ -0,0 +1,2281 @@
...
+
+            var argumentList = invocationExpression.ArgumentList as ArgumentListSyntax;
+            if (argumentList?.Arguments.Count != 1) return;
+            var argument = argumentList.Arguments.First();
+            if (argument.Expression as LiteralExpressionSyntax == null) return;
```
giggio: Use `IsKind`.

giggio:
I have found that code coverage is very low for `ValidateColorAnalyzer`. Please verify what is not being used and remove it. If you run `.\test.ps1` from root, you will see that coverage is at 31% (look at `log/codecoverage/index.htm`).

PR #443, opened 2015-07-29
Fix #424 forTaskNameAsyncAnalyzer with new or overrides modifiers

### LINKED ISSUE #424: BUG in CC0061 when the method is a override from a base class
When I was writing a Owin Middleware, using Microsoft.Owin DLL, Code Cracker suggested me to rename the Invoke method, that is defined in the OwinMiddleware class, to InvokeAsync, but if I do so, the compilation obviously fails. I think this will also happen if it were an interface.

The code bellow will reproduce the problem.
```C#
public class MyMiddleware : OwinMiddleware
	{
                public MyMiddleware(OwinMiddleware next)
			: base(next) { }

		public override Task Invoke(IOwinContext context)
		{
			return Next.Invoke(context);
		}
	}
```

adraut:
Ensure async methods with the new or overrides modifier are not flagged to be renamed by the TaskNameAsyncAnalyzer.


giggio:
Great!

Issue #404, opened 2015-07-11
Unit test methods raises CC0091 - "Make <xxxx> method static"


OsirisTerje:
Unit test methods are written as instance methods, and code cracker raises CC0091 on them. This warning is erroneous.  
The analyzer should suppress this for unit test methods.

If it uses the test attributes it should cover both MSTest, NUnit, XUnit and Intellitest methods. 

Attributes:
MSTest: `TestMethod`
NUnit: `Test`, `TestCase`, `TestCaseSource`
XUnit: `Fact`, `Theory`

In addition the setup and teardown attributes (XUnit uses ctor's so no issue there)
MSTest: `ClassInitialize`, `ClassCleanup`, `TestInitialize`, `TestCleanup`
NUnit: (V2): `TestFixtureSetup`, `TestFixtureTeardown`, `SetUp`, `TearDown` (V3), `OneTimeSetUp`, `OneTimeTearDown`

Also, any method that has the word `Test` in it is to be ignored, as NUnit uses that, but only if the containing class has a `TestFixture` attribute, or if there are sibling methods with a `Test` attribute. More info on this behavior [here](http://www.nunit.org/index.php?p=test&r=2.6.4) and [here](http://www.nunit.org/index.php?p=testFixture&r=2.6.4). Examples: `TestThis`, `DoATestSomewhere`.

giggio:
Agreed, but we have to be more specific. Can you provide a more detailed description on how we detect such methods? For example, in Nunit, allow tests to start with the word `test`, and don't need an attribute (http://www.nunit.org/index.php?p=test&r=2.6.4). Which cases do we want to except?

OsirisTerje:
There are many different ways of naming test methods, so requiring a certain naming convention can be hard.  
But, all test methods do have attributes that are marking them:
MSTest:  TestMethod
NUnit:  Test,  TestCase,  TestCaseSource
XUnit:  Fact
In addition the setup and teardown attributes (XUnit uses ctor's so no issue there)
MSTest:  ClassInitialize, ClassCleanup TestInitialize, TestCleanup
NUnit:  (V2): TestFixtureSetup, TestFixtureTeardown, SetUp, TearDown  (V3)  OneTimeSetUp, OneTimeTearDown

giggio:
Will we support mbUnit or none uses that anymore?
[...]

PR #444, opened 2015-07-29
Enhancement for #404 - Unit test methods raises cc0091

adraut:
#404 Add checks for test methods in the MakeMethodStaticAnalyzer.  Another method was added to the extensions class, HasAnyAttribute, because using HasAttributeOnAncestorOrSelf seemed like overkill what this needs analyzer to test.

### DIFF
--- a/src/CSharp/CodeCracker/Design/MakeMethodStaticAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Design/MakeMethodStaticAnalyzer.cs
@@ -86,7 +86,7 @@ namespace CodeCracker.CSharp.Design
             // Test if the method has any known test framework's attribute.
             result = method.AttributeLists.HasAnyAttribute(AllTestFrameworksMethodAttributes.Value);
 
-            if (!result && methodSymbol.Name.Contains("Test"))
+            if (!result && methodSymbol.Name.IndexOf("Test", System.StringComparison.OrdinalIgnoreCase) >= 0)
             {
                 // Test if the containing class has any NUnit class attribute
                 result = methodSymbol.ContainingType.GetAttributes().Any(attribute => attribute.AttributeClass.Name == NUnitTestClassAttribute);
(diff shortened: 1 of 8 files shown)


src/CSharp/CodeCracker/Design/MakeMethodStaticAnalyzer.cs
```diff
@@ -67,8 +72,39 @@ private static void AnalyzeMethod(SyntaxNodeAnalysisContext context)
...
         }
+
+        private static bool IsTestMethod(MethodDeclarationSyntax method, IMethodSymbol methodSymbol)
+        {
+            var result = false;
+
+            // Test if the method has any known test framework's attribute.
+            result = method.AttributeLists.HasAnyAttribute(AllTestFrameworksMethodAttributes.Value);
+
+            if (!result && methodSymbol.Name.Contains("Test"))
```
giggio: should work independent of case. "test" should also be ignored.

src/CSharp/CodeCracker/Design/MakeMethodStaticAnalyzer.cs
```diff
@@ -67,8 +72,39 @@ private static void AnalyzeMethod(SyntaxNodeAnalysisContext context)
...
+            {
+                // Test if the containing class has any NUnit class attribute
+                result = methodSymbol.ContainingType.GetAttributes().Any(attribute => attribute.AttributeClass.Name == NUnitTestClassAttribute);
```
giggio: This does not work, because the attribute name is actually reported as `TestFixtureAttribute` (it is a symbol).
Please add tests that consider this. You may have to actually declare the attribute classes on the tests to get this to work. It is easy to manually test, simply run and add NUnit as a reference, and create a test class with the `TestFixture` attribute with a simple method. Like this:

````csharp
using NUnit.Framework;
[TestFixture]
class Bar
{
    void Testa() { }
}
````

giggio:
@adraut I made some comments, please verify.

adraut:
@giggio I updated the verify diagnostic and create project methods so that a metadata reference can be passed in to make the tests work correctly.  I also added a reference in the test project to NUnit.  

Before the symbol was not being resolved which is why the test was passing when testing for TestFixture.  Now it only passes when testing for TestFixtureAttribute.

The test for methods with the name Test is now case insensitive, too.

giggio:
@adraut You don't need to add a reference to NUnit. Please remove that, it is an unnecessary thing. We already have to remove the reference to Json.NET, because of the same reason. If you let this happen we might soon end up with lots of references to libs. They are not needed. What you need to do is simply declare the type that NUnit has, on your test case, something like this:

````csharp
namespace NUnit.Framework
{
    [Attribute]
    public class TestFixtureAttribute { }
}
[TestFixture]
class Bar
{
    void Testa() { }
}
````

Also you did not change the other names. Shouldn't `TestCleanup` be `TestCleanupAttribute`. And all the others?

adraut:
@giggio That never occurred to me to do for the testing.  That's nice and easy.  I'll get that taken care of.

Yes, I need to go update all of the other ones that possible end in Attribute.  I got focused on NUnit and forgot about them.

giggio:
Ok, let me know when you are done.

adraut:
@giggio I removed the NUnit reference.  The other test attributes do not need to end with the `Attribute` because it is only testing the syntax instead of the symbol so the type name isn't being used.  The NUnit class level attribute is the only being tested by the symbol type name.

giggio:
Done.
I just added 2 extras commits. One (8f8f7db) to deal with a case where people used the full attribute name, and another (e65ae64) to revert the metadata references.

PR #447, opened 2015-07-30
Fix #432. Disposable Instantiation on constructor call 

### LINKED ISSUE #432: CC0022 disposable intantiation on constructor call
CC0022 is being reported on code like this:

    public EntitiesController() : this(new EntitiesDataContext())
    {
    }

Should this be detected as not an issue? Or is it better that it is explicit the reason why it's not being disposed?

thorgeirk11:
Potental Fix for #432.

Changes
-
- Added two checks to not raise when used in 
 - ThisConstructorInitializer, ```Foo() : this(new DisposableObject())```
 - ObjectCreationExpression ```new Bar(new DisposableObject())```
- Expanded the Code Fix to also provide a fix for disposable object passed as arguments.

```CSharp
string.Format(string.Empty,new Disposable());

// Becomes:

using (var disposable = new Disposable())
{
    string.Format(string.Empty, disposable);
}
```

### DIFF
--- a/src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedAnalyzer.cs
@@ -38,6 +38,8 @@ namespace CodeCracker.CSharp.Usage
             if (objectCreation == null) return;
             if (objectCreation?.Parent is UsingStatementSyntax) return;
             if (objectCreation?.Parent is ReturnStatementSyntax) return;
+            if (objectCreation.FirstAncestorOfType<ConstructorInitializerSyntax>()?.Kind() == SyntaxKind.ThisConstructorInitializer) return;
+            if (objectCreation.FirstAncestorOfType<ObjectCreationExpressionSyntax>() != null) return;
 
             var semanticModel = context.SemanticModel;
             var type = semanticModel.GetSymbolInfo(objectCreation.Type).Symbol as INamedTypeSymbol;
--- a/src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedAnalyzer.cs
@@ -38,8 +38,8 @@ namespace CodeCracker.CSharp.Usage
             if (objectCreation == null) return;
             if (objectCreation?.Parent is UsingStatementSyntax) return;
             if (objectCreation?.Parent is ReturnStatementSyntax) return;
-            if (objectCreation.FirstAncestorOfType<ConstructorInitializerSyntax>()?.Kind() == SyntaxKind.ThisConstructorInitializer) return;
-            if (objectCreation.FirstAncestorOfType<ObjectCreationExpressionSyntax>() != null) return;
+            if (objectCreation.Ancestors().Any(i => i.IsKind(SyntaxKind.ThisConstructorInitializer))) return;
+            if (objectCreation.Ancestors().Any(i => i.IsKind(SyntaxKind.ObjectCreationExpression))) return;
 
             var semanticModel = context.SemanticModel;
             var type = semanticModel.GetSymbolInfo(objectCreation.Type).Symbol as INamedTypeSymbol;
--- a/src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedCodeFixProvider.cs
+++ b/src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedCodeFixProvider.cs
@@ -62,12 +62,63 @@ namespace CodeCracker.CSharp.Usage
                 var statement = (LocalDeclarationStatementSyntax)variableDeclaration.Parent;
                 newRoot = CreateRootWithUsing(root, statement, u => u.WithDeclaration(variableDeclaration));
             }
+            else if (objectCreation.Parent.IsKind(SyntaxKind.Argument))
+            {
+                var identifierName = GetIdentifierName(objectCreation);
+
+                var whiteSpace = SyntaxFactory.Whitespace(" ");
+                var variableDeclaration = SyntaxFactory.VariableDeclaration(SyntaxFactory.IdentifierName(@"var").WithTrailingTrivia(whiteSpace))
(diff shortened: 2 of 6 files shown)


giggio:
There is a problem that, if you already have a variable in context with the same name, it will end up generating a variable with the same name. I recommend you add a `1`, `2`, etc, to the name, as we are doing on the `IntroduceFieldFromConstructorCodeFixProvider` (see lines 54 and 55).

This example will generate a compiler error:

````csharp
        void Foo
        {
            var str = "";
            var memoryStream = "";
            string.Format(str, new MemoryStream());
        }
````

**Comment by giggio** [comment]
I finished checking the code, it seems fine, the only problem is the one I reported above.

**Comment by giggio** [comment]
When you are done let me know.

File: src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedAnalyzer.cs [hunk 1]
```diff
@@ -38,6 +38,8 @@ private static void AnalyzeObjectCreation(SyntaxNodeAnalysisContext context)
             if (objectCreation == null) return;
             if (objectCreation?.Parent is UsingStatementSyntax) return;
             if (objectCreation?.Parent is ReturnStatementSyntax) return;
+            if (objectCreation.FirstAncestorOfType<ConstructorInitializerSyntax>()?.Kind() == SyntaxKind.ThisConstructorInitializer) return;
+            if (objectCreation.FirstAncestorOfType<ObjectCreationExpressionSyntax>() != null) return;
```
thorgeirk11: Are these two calls too heavy for the initial checking ?

src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedCodeFixProvider.cs
```diff
@@ -62,6 +62,38 @@ public static SyntaxNode CreateUsing(SyntaxNode root, ObjectCreationExpressionSy
                 var statement = (LocalDeclarationStatementSyntax)variableDeclaration.Parent;
                 newRoot = CreateRootWithUsing(root, statement, u => u.WithDeclaration(variableDeclaration));
             }
+            else if (objectCreation.Parent.IsKind(SyntaxKind.Argument))
+            {
+                var identifierName = GetIdentifierName(objectCreation, semanticModel);
+
+                var whiteSpace = SyntaxFactory.Whitespace(" ");
```
thorgeirk11: Is there a cached version of this somewhere ?

src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedCodeFixProvider.cs
```diff
@@ -69,6 +101,38 @@ public static SyntaxNode CreateUsing(SyntaxNode root, ObjectCreationExpressionSy
...
+                                   where symbolIdentifierName != null && symbolIdentifierName.StartsWith(identifierName)
+                                   select symbolIdentifierName;
+
+            var identifierPostFix = 0;
+            while (confilctingNames.Any(p => p == identifierName + ++identifierPostFix)) { }
+            return identifierName + (identifierPostFix == 0 ? "" : identifierPostFix.ToString());
+        }
+
+        private static string LoverCaseFirstLetter(string name) => char.ToLowerInvariant(name[0]) + name.Substring(1);
+        public static bool IncludeOnlyPartsThatAreName(SymbolDisplayPart displayPart) =>
```
thorgeirk11: This function is a duplicate from the NameOfAnalyzer, should I just call it or is this ok ?

src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedAnalyzer.cs
giggio: @thorgeirk11 Reflection is always costly. But this type of code appears everywhere, on many analyzers. We are not measuring those yet, so I wouldn't worry at this moment. What I would recommend is that you check other faster things first, and this is already happening.
BTW, the two calls above code be faster using `IsKind`. :)

src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedCodeFixProvider.cs
giggio: @thorgeirk11 of what?

src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedCodeFixProvider.cs
giggio: @thorgeirk11 Move it to an extension, replace it on both analyzers, and change the name to be more expressive. Something like `IsName`.

src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedCodeFixProvider.cs
thorgeirk11:
```CSharp
var whiteSpace = SyntaxFactory.Whitespace(" ");

giggio: No. But I don't think you need it. Just add `.WithAdditionalAnnotations(Formatter.Annotation)`.

src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedCodeFixProvider.cs
jwooley: Also, perhaps it should be named "LowerCaseFirstLetter" instead of "LoverCaseFirstLetter". While we like love our code, that may be going a bit too far ;-)

thorgeirk11: Haha :+1: Will correct the spelling error

src/CSharp/CodeCracker/Usage/DisposableVariableNotDisposedCodeFixProvider.cs
thorgeirk11: Alright, Thanks, will do

thorgeirk11:
Alright I think I am done. :)

thorgeirk11:
@giggio I think I am done :+1: 
(Forgot to mention you in my previous message)

giggio:
Done. It took me a while as I was busy at work. I also fixed the problem for calling into `base` on 8b707e5.

PR #453, opened 2015-08-01
Fix #452 for both CSharp and VB with proper casting.

### LINKED ISSUE #452: CC0029 DisposablesShouldCallSuppressFinalizeAnalyzer throws InvalidCastException
I'm getting an error that says AD0001, The User Diagnostic Analyzer 'CodeCracker.CSharp.Usage.DisposablesShouldCallSuppressFinalizeAnalyzer' threw an exception of type 'System.InvalidCastException' with message 'Unable to cast object of type 'Microsoft.CodeAnalysis.CSharp.Syntax.ThisExpressionSyntax' to type 'Microsoft.CodeAnalysis.CSharp.Syntax.IdentifierNameSyntax'.'.

I can't tell which type it's analyzing when it blows up, but my public Dispose() methods look like this:
```C#
public void Dispose()
{
	this.Dispose(true);
	GC.SuppressFinalize(this);
}
```
I've had to disable rule CC0029 in a ruleset file into order to get rid of this error.

adraut:
Added an extra cast to both the C# and VB analyzers.  Also had to get the statements of this dispose for VB a bit differently.  It was always returning Nothing with the previous code because it was getting a declaration node, which does not include the statements, instead of a method block node that does contain the statements.

### DIFF
--- a/src/CSharp/CodeCracker/Usage/DisposablesShouldCallSuppressFinalizeAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Usage/DisposablesShouldCallSuppressFinalizeAnalyzer.cs
@@ -49,7 +49,8 @@ namespace CodeCracker.CSharp.Usage
                 {
                     var invocation = statement.Expression as InvocationExpressionSyntax;
                     var method = invocation?.Expression as MemberAccessExpressionSyntax;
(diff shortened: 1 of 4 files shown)


giggio:
Hey, that was fast! Cool!

Issue #352, opened 2015-05-21
Test fix all providers


giggio:
Most code fix providers are not having their fix all actions tested. Also, most of them use the `BatchFixer`. We need to write tests to these code fix providers.
This action is an umbrella action. We will only close it when every code fix provider has tests.

- [ ] InterfaceNameCodeFixProvider
- [ ] SealedAttributeCodeFixProvider
- [ ] StaticConstructorExceptionCodeFixProvider
- [ ] StaticConstructorExceptionCodeFixProvider (VB)
- [ ] UseConfigureAwaitFalseCodeFixProvider (VB)
- [ ] DisposablesShouldCallSuppressFinalizeCodeFixProvider (VB)
- [ ] TernaryOperatorCodeFixProviders (VB)
- [ ] RemovePrivateMethodNeverUsedCodeFixProvider (VB)
- [ ] CatchEmptyCodeFixProvider (VB)
- [ ] InterfaceNameCodeFixProvider (C#)
- [ ] MustInheritClassShouldNotHavePublicConstructorsCodeFixProvider (VB)
- [ ] MakeLocalVariableConstWhenPossibleCodeFixProvider (VB)
- [ ] UnusedParametersCodeFixProvider (VB)
- [ ] RemoveWhereWhenItIsPossibleCodeFixProvider (VB)
- [ ] ComputeExpressionCodeFixProvider (C#)
- [ ] ArgumentExceptionCodeFixProvider (C#)
- [ ] EmptyCatchBlockCodeFixProvider (VB)
- [ ] UnnecessaryParenthesisCodeFixProvider (C#)
- [ ] UnusedParametersCodeFixProvider (C#)
- [ ] SealedAttributeCodeFixProvider (C#)
- [ ] EmptyFinalizerCodeFixProvider (C#)
- [ ] StaticConstructorExceptionCodeFixProvider (C#)
- [ ] DisposablesShouldCallSuppressFinalizeCodeFixProvider (C#)
- [ ] ConvertToSwitchCodeFixProvider (C#)
- [ ] RemoveUnusedVariablesCodeFixProvider (C#)
- [ ] RemovePrivateMethodNeverUsedCodeFixProvider (C#)
- [ ] ConvertLambdaExpressionToMethodGroupFixProvider (C#)
- [ ] EmptyObjectInitializerCodeFixProvider (C#)
- [ ] UseInvokeMethodToFireEventCodeFixProvider (C#)
[...]

PR #454, opened 2015-08-01
#352 Add fix all tests for InterfaceNameCodeFixProvider

adraut:


giggio:
This is good, but I think we are missing a test where an interface reference another from different source files, and making sure that it is fixed correctly.

PR #455, opened 2015-08-01
#352 Add fix all tests for SealedAttributeCodeFixProvider

adraut:


giggio:
We need to try to break things on these tests. The problem with fix all is that when we change things that might interfere with each other. For example, when a code file has 2 classes, can the fix all fix both? Sometimes it only fixes one, the other stays the same. Or, if a code file references something that is changing, and it is itself being fixed, does it work?
I am merging this, as the tests are helpful, but try to break stuff harder. And maybe add some more tests for this one specifically.

PR #464, opened 2015-08-19
Fix #461 for both VB and C# by testing for parameter symbol.

### LINKED ISSUE #461: Erroneous CC0039 message
In this code:

	Private Sub AdjustSample(readBuffer() As Single, ByRef readIdx As Integer, writeBuffer() As Single, ByRef writeIdx As Integer)
		If mIsMuted Then
			For chIdx = 0 To mDestChannels - 1
				writeBuffer(writeIdx) = 0
				readIdx += 1
				writeIdx += 1
			Next
...

there is an erroneous warning "Don't concatenate 'readIdx' in a loop".

adraut:


giggio:
As I explained on #461, this does not repro. I tried the tests you added, they pass without needing to change `StringBuilderInLoopAnalyzer`, so the code already worked. As I said on 461, if there is a bug we need a unit test that shows it. I am closing the PR, if you find the repro case let us know we will reopen it immediately.

Issue #466, opened 2015-08-20
Bug on DisposableVariableNotDisposedAnalyzer (CC0022) when returning a disposable


giggio:
This should NOT raise a diagnostic:

````csharp
public IDisposable Foo()
{
    var disposable = new MemoryStream();
    return disposable;
}
````

If the method returned `MemoryStream`, the diagnostic is already not being raised. Like so:

````csharp
public MemoryStream Foo()
{
     var disposable = new MemoryStream();
     return disposable;
}
````

ImaginaryDevelopment:
how could I tell when this fix is in a release or available for download in a beta or something?

giggio:
It is already available on myget: https://www.myget.org/F/codecrackerbuild/
I will be releasing a new version later at the end of upcoming week.

PR #467, opened 2015-08-22
Fix for bug #466.  Remove signature return type matches actual return…

adraut:
… type assumption.


giggio:
@adraut I don't know why that check was there. Thanks! :)

PR #474, opened 2015-08-29
Fix #469 by checking for a handles clause.

### LINKED ISSUE #469: CC0068 Reported Incorrectly for VB Event Handlers
The following code causes CC0068 Method is not used to be reported.
```
Private Sub Form_Load(ByVal sender as Object, ByVal e as EventArgs) Handles Me.Load
End Sub
```

adraut:


giggio:
@adraut great, tks! Next time, please do not reformat the code, it makes it really hard to check out the differences.

Issue #397, opened 2015-06-30
Bug on analyzer CC0031 (UseInvokeMethodToFireEvent) on parameterized functor


monoman:
Needed to disable CC0031 on this code, because this fix can't be applied as ```TReturn``` is not constrained to a reference type so ```getter?.Invoke(result)``` would cause a syntax error.

``` csharp
public static TReturn GetAssemblyAttributeValue<T, TReturn>(
    this Assembly assembly, 
    Func<T, TReturn> getter) where T : Attribute
{
  var result = assembly.GetAssemblyAttribute<T>();
#pragma warning disable CC0031 // Use Invoke Method To Fire Event Analyzer
  return (result != null) ? getter(result) : default(TReturn);
#pragma warning restore CC0031 // Use Invoke Method To Fire Event Analyzer
}
```

Probably you need to obtain an ```ITypeParameterSymbol``` from the  return type on ```typeInfo``` and see if it has the ```HasReferenceTypeConstraint``` property set, if not it should return;

:cherries:

giggio:
Bug confirmed. @monoman sorry, I don't know how I missed this.

giggio:
Closed on #475.

PR #475, opened 2015-08-29
Fix bug #397 by also checking for a parameter symbol.

adraut:


giggio:
@adraut This would remove analysis from all parameters, which is not the way to go.
We need to check if the return type is a reference type, otherwise it will fail. See the description on item #397, it is well detailed.
I have added a test to make sure we keep this type of check. See: https://github.com/code-cracker/code-cracker/blob/master/test/CSharp/CodeCracker.Test/Design/UseInvokeMethodToFireEventTests.cs#L196

adraut:
@giggio I completely missed the importance of the reference return type check.  I have updated the logic to account for that now.  I am unsure if the order the checks are done in will provide the best performance, though.

giggio:
@adraut I think the new order will make things faster. Thanks! :+1:

PR #477, opened 2015-08-29
Fix #476 by adding a check for attributes suffixed with Attribute

### LINKED ISSUE #476: Add additional checks for determining if generated code
#260 addressed most of the cases, but missed check for attributes ending with the Attribute suffix.
For example, the following should also check for DebuggerNonUserCodeAttribute and GeneratedCodeAttribute:
`[public static bool IsGenerated(this SyntaxNode node) => node.HasAttributeOnAncestorOrSelf("DebuggerNonUserCode", "GeneratedCode");]`

This is required to fix #472

adraut:


giggio:
@adraut Great. The problem with CC0013 was that it was also missing the generated code check. I fixed it: c35cd33.

Issue #480, opened 2015-08-29
CC0013 (user ternary) rule should be more careful with nullable types (C#)


giggio:
This will incorrectly raise `CC0013`:

````csharp
DateTime? Bar()
{
    if (true)
    {
        return null;
    }
    else
    {
        return new DateTime();
    }
}
````

The resulting code would be:

````csharp
DateTime? Bar()
{
    return true ? null : new DateTime();
}
````

Which does not compile.

This is related to #468 (VB), but in VB it does compile, but produces incorrect results.

sharwell:
Your example result doesn't seem to match the original condition. Is there a bug where the `if` and `else` branches are reversed, or was that a typo?

giggio:
@sharwell Right, fixed.

thomaslevesque:
How do you intend to fix this? Do you want to not raise the warning in this case, or to change the code fix so that it generates valid code?

giggio:
@thomaslevesque I was thinking about not raising, but now that you raised the question, we could do this on the fix:

````csharp
DateTime? Bar()
{
    return true ? (DateTime?)null : new DateTime();
}
````

So, simply casting the reference to the method's return type.

thomaslevesque:
I started working on this, then I realized that @adraut had already submitted a PR...

While working on my fix I realized that the 2 code fix providers (for return & assignment) would benefit from having a common abstract base class; there's a lot of logic that could be factorized. Also, it would probably be cleaner to use `SyntaxFactory.ConditionalExpression` rather than `SyntaxFactory.ParseStatement`.

adraut:
Sounds like @thomaslevesque has a better solution than mine.  I also had the thought that the proposed solution for this issue would better than what was proposed for #468.  I think I should close both PR's.

thomaslevesque:
I submitted my own PR as well, since it was almost done.

I also included the changes mentioned in my previous comments.

giggio:
That is why we ask people to comment on the issue, saying that they want to work on it, so we don't have two people working on the same issue.
I would usually select the first one and close the second one. But it seems @adraut is fine with we looking into @thomaslevesque 's. Is that ok, @adraut?

adraut:
@giggio yes that is fine.  I would rather the better solution be used.

PR #483, opened 2015-08-31
Fix #468 by checking for mixed reference and value type return types.

### LINKED ISSUE #468: CC0013 (user ternary) rule should be more careful with nullable types. (VB)
for example, this:
    If value Is DBNull.Value Then
                    Return Nothing
    Else
                    Return CDate(value)
    End If
not equals to:
Return If(value Is DBNull.Value, Nothing, CDate(value))

first will return Nothing, but second will return DateTime.MinValue.

adraut:


giggio:
@adraut Assignment is not working, it will still raise a diagnostic:

````vb
    Function foo() As Date?
        Dim d As DateTime?
        If True Then
            d = Nothing
        Else
            d = CDate(Nothing)
        End If
    End Function
````

adraut:
@giggio  I think this PR should be closed.  #468 should be fix the same way #480 was fixed.

giggio:
Ok.

PR #487, opened 2015-08-31
Fix #480

### LINKED ISSUE #480: CC0013 (user ternary) rule should be more careful with nullable types (C#)

thomaslevesque:
Fix #480 

Explicitly cast null to nullable type in conditional expression.

Extract common functionality of the 2 code fix providers into an abstract
base class.


giggio:
@thomaslevesque Good job!

Issue #151, opened 2014-12-10
Remove async and return task directly


giggio:
Sometimes an async method might only do some some sync work and then at the end await on another async method. Like this:

````csharp
async Task FooAsync()
{
    DoSomething();
    await SomethingElseAsync();
}
````

To optimize the generated code we could simply return the task from `SomethingElseAsync`. So the previous code becomes:

````csharp
Task FooAsync()
{
    DoSomething();
    return SomethingElseAsync();
}
````

This is the decompiled C# generated on the first case (optmized, release settings):

````csharp
		private Task FooAsync()
		{
			Program.<FooAsync>d__1 <FooAsync>d__;
			<FooAsync>d__.<>4__this = this;
			<FooAsync>d__.<>t__builder = AsyncTaskMethodBuilder.Create();
			<FooAsync>d__.<>1__state = -1;
			AsyncTaskMethodBuilder <>t__builder = <FooAsync>d__.<>t__builder;
			<>t__builder.Start<Program.<FooAsync>d__1>(ref <FooAsync>d__);
			return <FooAsync>d__.<>t__builder.Task;
		}
````

Where in the second case it stays pretty much the same, there is no need to generate the state machine. I was actually expecting the compiler to be smarter, but it is not.

Category: `Performance`
Severity: `Info`
Diagnostic id: `CC0051`

giggio:
@code-cracker/owners This has been discusses a long time ago, I am moving it to `Ready`.

thorgeirk11:
Is it ok if I try to tackle this ?

giggio:
Go for it.
Remember to fork from the `vnext` branch, not from `master`.

thorgeirk11:
Is there a place where I can change the info displayed when navigated to the error page for the this diagnostic id?

That is to add an error page for https://code-cracker.github.io/diagnostics/CC0051.html

giggio:
Worked on #497.

PR #497, opened 2015-09-05
Return task instead of await

thorgeirk11:
This PR introduces the ReturnTaskInsteadOfAwaitAnalyzer.
Discussed in #151. 
I also added 26 Unit tests to cover both the analyzer and CodeFix. 

Please let me know what you think :)

### DIFF
--- a/src/CSharp/CodeCracker/CodeCracker.csproj
+++ b/src/CSharp/CodeCracker/CodeCracker.csproj
@@ -47,6 +47,7 @@
     <Compile Include="Design\MakeMethodStaticCodeFixProvider.cs" />
     <Compile Include="Extensions\CSharpAnalyzerExtensions.cs" />
     <Compile Include="Extensions\CSharpGeneratedCodeAnalysisExtensions.cs" />
+    <Compile Include="Performance\ReturnTaskInsteadOfAwaitAnalyzer.cs" />
     <Compile Include="Performance\UseStaticRegexIsMatchAnalyzer.cs" />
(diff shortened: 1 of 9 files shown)


giggio:
@thorgeirk11 Looks pretty good. But you are not using the resources for globalization. Check out `NameOfAnalyzer` and code fix for an example.

thorgeirk11:
@giggio Ah thanks, I wasn't sure how to use it. 
I have pushed a fix for that, so now it uses the resources correctly. :+1:

giggio:
@thorgeirk11 I am getting a null reference exception on this code:

````csharp
        public static async Task FooAsync()
        {
            await Task.Delay(0);
        }
````

thorgeirk11:
Will look into it when I get home ! :+1:

thorgeirk11:
This seems to work in a unit test. Could I be missing something ?
image

giggio:
@thorgeirk11 I am not sure why I got that problem. I reset the Rolsyn VS instance and the problem went away.

giggio:
@thorgeirk11 Actually, I got the problem to show up again. It does not show on your test because it only triggered on incomplete methods, the problem was that `methodDecl.Body` was null while the user typed.
This would trigger it:

````csharp
public static async Task FooAsync()
````

(without the body)

Also, on your test, you are not wrapping the method on a class/struct, so there would be other C# errors. This happened on other tests of yours.

It is a simple fix, I fixed it myself on f4b15d4. Also see the other small changes I did.

Issue #489, opened 2015-09-01
Prefer "Count" to "Count()"


giggio:
If it is an `IList` or some other object that has `Count` as a property, don't use the Linq extension method.

So this:

````csharp
var count = myList.Count();
````

Becomes:

````csharp
var count = myList.Count;
````

**Important**:
`Count` and `Count()` have to return the same type.
`Count()` has to be an extension method.

Diagnostic id: `CC0098`
Severity: Warning
Category: Performance

thorgeirk11:
Can I take this one ?

giggio:
Closed on #567.

Issue #490, opened 2015-09-01
Prefer "Any" to "Count() > 0"


giggio:
If you find a call to `Count() > 0` and the object supports `Any()`, change it `Any()`.
If you find a call to `Count(predicate) > 0` and the object supports `Any(<predicate type>)`, change it `Any(predicate)`.

So this:

````csharp
var any = myEnumerable.Count() > 0;
````

Becomes:

````csharp
var any = myEnumerable.Any();
````

And:

````csharp
var any = myEnumerable.Count(i => i > 3) > 0;
````

Becomes:

````csharp
var any = myEnumerable.Any(i => i > 3);
````

Diagnostic id: `CC0099`
Severity: Information
Category: Style

akamud:
I would like to work on this

giggio:
Go for it!

giggio:
@akamud Remember to branch from vnext, and to pull into it. I have added you to the collaborators team, so I can assign the issue to you.

akamud:
Will do.

giggio:
Closed on #528

PR #504, opened 2015-09-08
Fix issue #489

pmercer:
Updated uses of .Count() with either .Count or .Length accordingly.


giggio:
This was not actually a fix for #489 or #490, those are analyzer issues, we have to build analyzers that do that automatically.
But hey, thanks! :)

Issue #382, opened 2015-06-14
Extract Class to a New File


carloscds:
Extracting a class to an external file, since this is not the only class in the file. The extracted class may not have the same name as the source file, if it occurs, change the file name including a number at the end.

The class must be created in the same folder and namespace.

Category: `Refactoring`
Diagnostic Id: `CC0096`
Severity: `Hidden`

sharwell:
See the following for potential implementation assistance:

* DotNetAnalyzers/StyleCopAnalyzers#954
* DotNetAnalyzers/StyleCopAnalyzers#953

carloscds:
@sharwell Thanks.

giggio:
Closed on #505

giggio:
This is a cool one. And it is done now!
:shipit:

PR #505, opened 2015-09-09
Issue #382

carloscds:
Fix issue #382

### DIFF
--- a/src/CSharp/CodeCracker/CodeCracker.csproj
+++ b/src/CSharp/CodeCracker/CodeCracker.csproj
@@ -54,12 +54,14 @@
     <Compile Include="Performance\StringBuilderInLoopAnalyzer.cs" />
     <Compile Include="Performance\UseStaticRegexIsMatchCodeFixProvider.cs" />
     <Compile Include="Refactoring\AddBracesToSwitchSectionsAnalyzer.cs" />
+    <Compile Include="Refactoring\ExtractClassToFileCodeFixProvider.cs" />
     <Compile Include="Refactoring\AddBracesToSwitchSectionsCodeFixProvider.cs" />
     <Compile Include="Refactoring\AllowMembersOrderingAnalyzer.cs" />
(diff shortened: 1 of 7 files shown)


giggio:
@carloscds You need to use resources. See the nameof analyzer and fix.

giggio:
@carloscds 

A few issues:

Analyzer does not work if classes are not inside of a namespace.

This code will have several problems:

````csharp
using System;
namespace ConsoleApplication3
{
    using System.Threading.Tasks;

    public class Class1
    {
        DateTime? D() => null;
        async Task Go() { }
    }
    public class Class2
    {
        DateTime? D() => null;
        async Task Go()
        { 
        }

    }
}
namespace ConsoleApplication3.Foo
{
    class Class3
    {
    }
    class Class4
    {
    }
}
````

* Class3 anc Class4 will not raise the diagnostic, and I think it is throwing.
* If you extract Class1 or Class2 the `using System.Threading.Tasks;` will be moved on the new file from inside the namespace declaration to the top of the file.

carloscds:
@giggio Do you have a class outsite namespace ?

giggio:
@carloscds A simple test like this will expose the bug:

````csharp
class Class1 { }
class Class2 { }
````

carloscds:
@giggio My question is: is there a project that has a class outside namespace ?

giggio:
@carloscds I have seen code like this. I don't think it is a good idea to ignore. It would be better to use the compilation root as the base, not the namespace. What do you think?

carloscds:
@giggio For now I think that's ok. Maybe we can improve it in future.

giggio:
There are other errors, did you see?

carloscds:
@giggio I update my PR, did you see ?

carloscds:
@giggio PR updated!

giggio:
@carloscds Still not working with classes under different namespaces. Needs to update analyzer.

carloscds:
@giggio It's working fine. I have tested with two separate namespaces.

giggio:
Now it is perfect!

PR #513, opened 2015-09-16
Fix #468 by casting to nullable types as necessary.

### LINKED ISSUE #468: CC0013 (user ternary) rule should be more careful with nullable types. (VB)

adraut:
The correct code is being generated by the code fix provider, but I can not get the tests using `VerifyBasicFixAsync` to pass because it throws an error of
```
Fix introduced new compiler diagnostics:
Test0.vb(5) : error BC30002: Type 'Integer?' is not defined.
```
If I switch them to use `VerifyBasicFixAllAsync` they pass just fine.  Is there a known bug in `VerifyBasicFixAsync` or `VerifyBasicFixAllAsync`?

### DIFF
--- a/src/VisualBasic/CodeCracker/Style/TernaryOperatorCodeFixProviders.vb
+++ b/src/VisualBasic/CodeCracker/Style/TernaryOperatorCodeFixProviders.vb
@@ -8,9 +8,25 @@ Imports Microsoft.CodeAnalysis.VisualBasic
 Imports Microsoft.CodeAnalysis.VisualBasic.Syntax
 
 Namespace Style
+
+    Public MustInherit Class TernaryOperatorCodeFixProviderBase
+        Inherits CodeFixProvider
+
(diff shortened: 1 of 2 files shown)


giggio:
@adraut You have one test failing during build. Please verify.

giggio:
@adraut I just saw the PR message. I am now aware of any bugs on those methods...

adraut:
@giggio I've stepped through the visual basic fix verification methods and still haven't been able to figure out why `VerifyBasicFixAsync` fails but the `VerifyBasicFixAllAsync` succeeds.  Any help would be appreciated.

giggio:
@adraut I am away from home, I will only be able to spend time on it next weekend.

carloscds:
@adraut Can you try to update your branch with CodeCracker/master and commit again ?

adraut:
Close to try again.

giggio:
I was going to look at it just now. If you are still having problems let us know, I will be able to look into it this weekend. Just update it to HEAD so it is easier to check.

Issue #514, opened 2015-09-17
CC0017 Change to auto property fix all not working


thorgeirk11:
The code fix for CC0017 doesn't change all instances.

If the code fix is run on a document containing the following code:
```CSharp
private int x;
private int y;
public int X
{
    get { return x; }
    set { x = value; }
}

public int Y
{
    get { return y; }
    set { y = value; }
}
```
It results in the following:
```CSharp
private int x;
public int X
{
    get { return x; }
    set { x = value; }
}
public int Y { get; set; }
```

But of course it should be: 
```CSharp
public int X { get; set; }
public int Y { get; set; }
```

I found this out when I was working on #500 and #512 but I am unsure how to fix it so I decided to just make a Issue out of it.

giggio:
Bug confirmed. This should be fixed on the `master` branch. It is up for grabs.

giggio:
BTW, there is already an ignored test on the code for this specific bug.

adraut:
Almost seems like a threading issue.  Sometimes the result is
```C#
private int y;

public int X { get; set; }

public int Y
{
    get { return y; }
    set { y = value; }
}
```
and other times it's
```C#
private int x;

public int X
{
    get { return x; }
    set { x = value; }
}

public int Y { get; set; }
```

The fixer does request the root a few times throughout the process, but I've added debug writes in the fix process to see what the document looks like at each step and it does not appear that one fix for a property is interacting with another.

Looks like it could possibly be due to https://github.com/dotnet/roslyn/issues/4416.  It appears that maybe some trivia is being seen as changed by multiple fixes and only one gets merged.

It's very easy to see with the existing test case that was previously being skipped.
```
private int x = 42, y;

public int X
{
    get { return x; }
    set { x = value; }
}

public int Y
{
    get { return y; }
    set { y = value; }
}
```

Will randomly return
```
private int y;

public int X { get; set; } = 42;

public int Y
{
    get { return y; }
    set { y = value; }
}
```

Or 
```
private int x = 42;

public int X
{
    get { return x; }
    set { x = value; }
}

public int Y { get; set; }
```

giggio:
@adraut The problem is that the batch fixer only solves non conflicting changes, and this is creating a conflict. It does not matter that it is not deterministic, it would only fail in a deterministic way, but it would still fail.
What is needed to fix this is work in a way that all changes are merged together in an intelligent way. This demands code that we write, the batch fixer is unable to conciliate this problems.
[...]

Issue #512, opened 2015-09-15
CC0017 Change to auto property codefix removes multiple variable declaration.


thorgeirk11:
While I was working on #500 I encountered another issue:
CodeFix removes multiple variable declaration.
---
The following:
```CSharp
public int x, y;
public int X
{
    get { return x; }
    set { x = value; }
}
```
Get transformed into
```CSharp
public int X { get; set; }
```
The `y` gets removed.

thorgeirk11:
I will fix this along with #500 :+1:

giggio:
Ok, go for it!

giggio:
Closed on #515

PR #515, opened 2015-09-17
CC0017 ignores field assignments

### LINKED ISSUE #500: CC0017 Change to auto property ignores field assignment.
Ignores field assignments.
---

The code fix for the CC0017 ignores when the field has an value set to it.

The following code:
```CSharp
private int myVar = 42;

public int MyProperty
{
    get { return myVar; }
    set { myVar = value; }
}
```

Gets fixed to the following:
```CSharp
public int MyProperty { get; set; }
```

But should be changed to:
```CSharp
public int MyProperty { get; set; } = 42;
```

---

Note that if the value is the default value this can be ignored.
So the following is totaly valid:
```CSharp
private int myVar = 0;

public int MyProperty
{
    get { return myVar; }
    set { myVar = value; }
}
// Transformed into:
public int MyProperty { get; set; }
```

----

For language level below 6 this case should be ignored by the analyser. Since setting the field is an valid way of assigning the property a default value.

thorgeirk11:
This pull request fixes #500 and #512. 

I included a unit test for #514. I was unsure how to fix that issue so I will just leave it in.

Please feel free to comment on this.

### DIFF
--- a/src/CSharp/CodeCracker/Style/SwitchToAutoPropCodeFixProvider.cs
+++ b/src/CSharp/CodeCracker/Style/SwitchToAutoPropCodeFixProvider.cs
@@ -52,10 +52,7 @@ namespace CodeCracker.CSharp.Style
             returnIdentifier = root.GetCurrentNode(returnIdentifier);
             returnIdentifierSymbol = semanticModel.GetSymbolInfo(returnIdentifier).Symbol;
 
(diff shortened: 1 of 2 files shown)


carloscds:
@thorgeirk11 Good job!

giggio:
This should had to go on the master branch. I will cherry pick to it. Bug fixes are all going to master, new stuff goes to vnext.

PR #527, opened 2015-09-29
Fixes incorrect help links from Analyzers

akamud:
I noticed some analyzers had the wrong Help link Uri, so I fixed them.

Since this is a simple fix I used the `master` branch. If this should be on `vnext` just let me know and I'll redo.


giggio:
Perfect! Thanks!

PR #528, opened 2015-09-30
Add PreferAnyToCountGreaterThanZero Analyzer, closes #490

### LINKED ISSUE #490: Prefer "Any" to "Count() > 0"

akamud:
Implemented this code fix for `Count() > 0` and `Count > 0` in C# and VB

### DIFF
--- a/src/CSharp/CodeCracker/CodeCracker.csproj
+++ b/src/CSharp/CodeCracker/CodeCracker.csproj
@@ -56,6 +56,8 @@
     <Compile Include="Performance\StringBuilderInLoopAnalyzer.cs" />
     <Compile Include="Performance\UseStaticRegexIsMatchCodeFixProvider.cs" />
     <Compile Include="Refactoring\AddBracesToSwitchSectionsAnalyzer.cs" />
+    <Compile Include="Style\PreferAnyToCountGreaterThanZeroCodeFixProvider.cs" />
+    <Compile Include="Style\PreferAnyToCountGreaterThanZeroAnalyzer.cs" />
     <Compile Include="Refactoring\ExtractClassToFileCodeFixProvider.cs" />
     <Compile Include="Refactoring\AddBracesToSwitchSectionsCodeFixProvider.cs" />
     <Compile Include="Refactoring\AllowMembersOrderingAnalyzer.cs" />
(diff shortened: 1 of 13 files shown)


giggio:
@akamud It is accepting `xs.Count() > 1`. It should only work with `> 0`. Please fix that on VB and C#.
Also, I noticed you are not using the Count Property, on the method invocation. I know it was not specified on #490, but I think it makes sense to have it as well, what do you think? Can you add it?

akamud:
Oops, forgot to check for 0 only.

I actually skipped Count (property) because of [this analyzer](https://github.com/code-cracker/code-cracker/issues/489) that is being implemented. What do you think, is Any() better than Count > 0?

giggio:
@akamud I think you can still do `Count`, even though #489 is there.
I noticed that for a Collection, Any is slightly inefficient than checking the count. See:
Any:
https://github.com/dotnet/corefx/blob/master/src/System.Linq/src/System/Linq/Enumerable.cs#L1874
Count:
https://github.com/dotnet/corefx/blob/master/src/System.Linq/src/System/Linq/Enumerable.cs#L1905
In `Any` it always gets the enumerator, and then it enumerates. That is more expensive than checking for `Count`. But the readability of `Count` is better. And `Count()` is special cased to deal with collections.
Anyway, I'd offer it to both `Count()` and `Count`.

akamud:
I fixed the C# and VB analyzers and rebased.  
Anything else missing?

giggio:
@akamud Got a few conditions that needed fixing and fixed them. Check out 80c14a1.

akamud:
Thanks @giggio :+1:

PR #530, opened 2015-10-03
Fix for #468 ternary code fix with nullables

adraut:
The single fix tests that call VerifyBasicFixAsync tests still do not pass.  If I change them to call VerifyBasicFixAllAsync instead it works.  Makes no sense to me why Integer? would ever be undefined.

### DIFF
--- a/test/VisualBasic/CodeCracker.Test/Style/TernaryOperatorTests.vb
+++ b/test/VisualBasic/CodeCracker.Test/Style/TernaryOperatorTests.vb
@@ -180,47 +180,54 @@ End Namespace"
     End Function
 
     <Fact>
-    Public Async Function WhenTernaryWithObjectDoesApplyFix() As Task
+    Public Async Function WhenUsingIfAndElseWithNullableValueTypeAssignmentChangeToTernaryFix() As Task
         Const source = "
-Class MyCustomer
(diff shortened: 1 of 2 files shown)


test/VisualBasic/CodeCracker.Test/Style/TernaryOperatorTests.vb
```diff
@@ -180,47 +180,54 @@ End Namespace"
...
-Public Class ExcelLineRecordClass
-    Public Property LineNumber As Integer
-    Public Property imported As Boolean
-End Class
-Class Tester
-    Private Sub Test()
-        Dim ExcelRecord As New ExcelLineRecordClass
-        Dim lCell As New MyCustomer
-        If lCell Is Nothing Then
-            ExcelRecord.imported = False
```
giggio: Why was this test removed?

giggio:
@adraut I am not sure `Integer?` is being declared as undefined. Anyway, just pass `allowNewCompilerDiagnostics:=True` and it will work:

````vb
Await VerifyBasicFixAsync(source, fix, allowNewCompilerDiagnostics:=True)
````

But there is another issue here, you seem to have a merge problem, you have deleted code from a previous bug fix. Please verify.

Issue #533, opened 2015-10-03
UnusedParametersCodeFixProvider will crash when it is trying to remove ParamArray


paul1956:
For the following Sub
```
    Private Sub PrintSummary(ByRef Worksheet As IXLWorksheet, ByVal colData As ColDataCollection, ByVal ParamArray levels() As Integer)
        Dim ExcelRowIndex As Integer = FindExcelRow(Worksheet, colData(0).Value, False)
        For colDataIndex As Integer = 1 To colData.Count - 1
            Worksheet.Cell(ExcelRowIndex, colDataIndex + 2).Value = colData(colDataIndex).Value
        Next
    End Sub
```
with references that look like
```
PrintSummary(Worksheet, Total.ColDataList)
```
 Function RemovePArameterAsync needs to change from
```
Dim newArguments As ArgumentListSyntax = arguments.WithArguments(arguments.Arguments.RemoveAt(parameterPosition))
```
```
If arguments.Arguments.Count > parameterPosition Then
     Dim newArguments As ArgumentListSyntax = arguments.WithArguments(arguments.Arguments.RemoveAt(parameterPosition))
    replacingArgs.Add(arguments, newArguments)
End If
 ```

giggio:
Bug confirmed. It is up for grabs. Are you going to work on it, @paul1956?

paul1956:
I don't know how to work on GitHub, the two line change I posted fixes the bug. If you want to point to the process to "works on it" I will.

giggio:
@paul1956 There was a problem on the way params was being handled. It had to remove all arguments, and it was only removing one. I fixed that and the exception is not happening anymore.
The problem also ocurred with C#.
See #535 and specially commit 873afb1. I also reenabled the analyzer and fixed the fix all, which was not working.

paul1956:
The case that was causing my crash was when there are 0 parram array arguments calling the function.

PR #535, opened 2015-10-04
Fix unused parameters

giggio:
Two updates:

Enable params with UnusedParameters (#533)
Also reenable the analyzer for VB

Fix all in UnusedParameters (#534)
Also update VB analyzer to send props to code fix and update code fix making it more effient when reporting the fix


giggio:
@paul1956 It does not crash anymore. See:
notcrashing

paul1956:
Try Go(1,2) unless you changed something since I last checked it will crash. The problem is when you have a caller that does not pass anything to ParamArray. ParamArray is similar to Optional in that you can leave it completely out.

giggio:
@paul1956 Ok, got it to reproduce. I will reopen that bug.

Issue #537, opened 2015-10-04
CC0001 Is raised on multiple variable declarations.


thorgeirk11:
The following code raises CC0001 
```CSharp
public void Foo()
{
    string foo = "", otherFoo = "";
}
```
However the code fix introduces a compile error:
``` Error	CS0819	Implicitly-typed variables cannot have multiple declarators```

giggio:
Bug confirmed.

giggio:
Closed on #549.

PR #549, opened 2015-10-11
Fixed issue #537 by creating multiple variable declarations rather th…

Shiney:
…an a single declaration.

I added a test for this issue directly, and also added a test that the CodeFix doesn't destroy the trivia.

### DIFF
--- a/src/CSharp/CodeCracker/Style/AlwaysUseVarCodeFixProvider.cs
+++ b/src/CSharp/CodeCracker/Style/AlwaysUseVarCodeFixProvider.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.CodeAnalysis;
+﻿using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CodeActions;
 using Microsoft.CodeAnalysis.CodeFixes;
 using Microsoft.CodeAnalysis.CSharp;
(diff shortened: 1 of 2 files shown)


Shiney:
#537

thorgeirk11:
I am not sure we can safely use var for this case. See the following example:

```CSharp
class Animal
{
    public string Foo() { return "Animal"; }
}
class Cat : Animal
{
    public new string Foo() { return "Cat"; }
}
class Dog : Animal
{
    public new string Foo() { return "Dog"; }
}
public void Bar()
{
    Animal dog = new Dog(), cat = new Cat();
    Console.WriteLine(dog.Foo()); // Prints out Animal.
    Console.WriteLine(cat.Foo()); // Prints out Animal.
}
```

However when refactored:
```CSharp
public static void Bar()
{
    var dog = new Dog();
    var cat = new Cat();
    Console.WriteLine(dog.Foo()); // Prints out dog.
    Console.WriteLine(cat.Foo()); // Prints out cat.
}
```

What do you think @code-cracker/collaborators ?

akamud:
Indeed. We could check if the left side is a base type of the right side before suggesting `var`.

Shiney:
foreach (var variable in variableDeclaration.Variables)
            {
                if (variable.Initializer == null) return;
                var conversion = semanticModel.ClassifyConversion(variable.Initializer.Value, variableType);
                if (!conversion.IsIdentity) return;
            }

This code is in the analyser and should be stopping that example. I'm on my phone so I have not checked there is a test for that.

Also that issue with the code fix is not related to the multiple declarations. If it's an issue then it would be an issue with single declarations too.

FYI resharper does not seem to support this refactoring on multiple declarations. Maybe it's not that valuable as it increases the number of lines of code.

Thanks for committing this.

PR #556, opened 2015-10-17
Fixes #544 by ignoring assignment of fields in Lambda in constructor

### LINKED ISSUE #544: BUG on ReadonlyFieldAnalyzer (CC0052) with assignment in Func<T> in constructor
With this code :
```
internal class Test
{
    private string _value;

    public Test()
    {
        Func<string> factory = () => _value ?? (_value = "Hello");
    }
}
```

the ReadonlyFieldAnalyzer shows a suggestion for the field ```_value```, but the field ca be assigned outside of the constructor (when invoking the Func).

mgrosperrin:
This PR fixes #544 by checking if the assignment of a field in a constructor is done in a lambda.


carloscds:
@mgrosperrin Thanks for PR!

PR #558, opened 2015-10-17
CS0058 & CS0059 code-fix

baks:
Fixes `CS0058` and `CS0059` in #381.

Implementation with tests for `CS0058` and `CS0059` compiler error numbers.

Additionally I added checking if reason for inconsistent accessibility is in accessibility domain instead of actual accessibility modifiers of type in cause of nested types.

I created a `CreateDocument` method overload without `LanguageVersion` for `VisualBasic` in `DiagnosticVerifier.cs` because I wanted to avoid draging dependency to `..CodeAnalysis.VisualBasic` for `CodeCracker.Test` `C#` project


giggio:
@baks Really good. Only 2 to go!

PR #563, opened 2015-10-23
bug fixed

### LINKED ISSUE #547: CallExtensionMethodAsExtension (CC0026) throws NullReferenceException on expression bodied method statement
Here is a test method for reproducing this behavior:
```C#
  [Fact]
  public async Task WhenCallExtensionMethodAfterArrowExpressionAsStaticMethodTriggerAFix()
  {
      const string source = @"
              using System.Linq;
              namespace ConsoleApplication1
              {
                  public class Foo
                  {
                      static int[] source = new int[] { 1, 2, 3 };
                      public static void Bar() => Enumerable.Any(source);
                  }
              }";

      var expected = new DiagnosticResult
      {
          Id = DiagnosticId.CallExtensionMethodAsExtension.ToDiagnosticId(),
          Message = "Do not call 'Any' method of class 'Enumerable' as a static method",
          Severity = DiagnosticSeverity.Info,
          Locations = new[] { new DiagnosticResultLocation("Test0.cs", 8, 57) }
      };

      await VerifyCSharpDiagnosticAsync(source, expected);
  }
```

carloscds:
Fix #547


giggio:
@carloscds The analyzer should still raise a diagnostic for arrow functions. The way you implemented you fixed the bug, but the analyzer is not working as expected. The test on #547 is what we should be checking. Also, a fix test would be good too.

carloscds:
@giggio Bug is about nullexception, not about behavior of analyzer.

PR #567, opened 2015-10-26
Change Count() to Count

### LINKED ISSUE #489: Prefer "Count" to "Count()"

carloscds:
Fix #489.


giggio:
@carloscds The PR is fine, but I caught a bug. If you apply the code fix here it will create a compile error:

````csharp
class Program
{
    static void Main(string[] args)
    {
        var b = new Bar();
        Foo(b.Count());
    }
    static void Foo(int i)
    {
    }

    class Bar : List<int>
    {
        public string Count { get; set; }
    }
}
````

You need to check that the property and method have the same return type.

carloscds:
@giggio Bug fixed.

giggio:
@carloscds now failing for this, throwing nullreferenceexception:

````csharp
class Program
{
    static void Main(string[] args)
    {
        var b = new Bar();
        Foo(b.Count());
    }
    static void Foo(int i)
    {
    }

    class Bar : List<int>
    {
        public string Count;
    }
}
````

carloscds:
@giggio Fixed again!

PR #573, opened 2015-10-29
CS0060 and CS0061 Inconsistent accessibility code fix

baks:
Fixes `CS0060` and `CS0061` in #381.

Implementation with tests for `CS0060` and `CS0061` inconsistent accessibility compiler errors.

It looks like we finally have fix for every compiler error related to inconsistent accessibility :smiley:


giggio:
@baks I caught an interesting bug. If you have this code:

````csharp
class Program
{
    static void Main(string[] args)
    {
    }
}
public class Foo
{
    protected class Bar : Program
    {
    }
}
````

If you apply the fix from the `Bar` class, the `Program` class will become `protected`, and that would be illegal. It should be `public`.

baks:
@giggio, thank you for catching this. I've fixed it.

Also I've made small refactoring in class/method names, files location.

giggio:
@baks Would you pleas fix the conflicts?

baks:
@giggio, thank you for notifying me. I didn't notice that the conflicts have arisen.

PR #584, opened 2015-11-05
Fix failing ValidateColor tests in some cultures

thomaslevesque:
Some tests in ValidateColorTests were failing in French culture. This is because `"200,100,50"` is a valid color in English, but not in French, because the list separator is different. See the comments [here](https://github.com/code-cracker/code-cracker/issues/1#issuecomment-154075860) for details.

I made a few changes to the `ChangeCulture` class:
- set the `DefaultThreadCurrentCulture` and `DefaultThreadCurrentUICulture` properties so that the new culture also applies to other threads in async code
- save the original values to be able to restore them later

### DIFF
--- a/test/Common/CodeCracker.Test.Common/ChangeCulture.cs
+++ b/test/Common/CodeCracker.Test.Common/ChangeCulture.cs
@@ -1,23 +1,37 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+using System.Globalization;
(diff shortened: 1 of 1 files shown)


giggio:
@thomaslevesque I just pulled the changes into master as well.
Did this solve the test problem you were having on the machines with non en-us culture?

thomaslevesque:
> Did this solve the test problem you were having on the machines with non en-us culture?

Yes, it did. Setting DefaultThreadCurrentCulture causes new threads to use the specified culture, so it works with async code as well.

PR #588, opened 2015-11-06
Fixes #587 - stripping out non-selected catch blocks. Only remove try…

### LINKED ISSUE #587: Remove empty Catch Block too aggressive
If code has multiple catch blocks in a single try, when selecting the fix to remove empty catch block, it removes all catch blocks not just the one requested. Sample test:

  public void foo()
    {
        try
        {
            Console.WriteLine("foo");
        }
        catch (ArgumentException aex)
        {  
            
                   
        }
        catch 
        {
            throw;
        }
    }

Should return 
  public void foo()
    {
        try
        {
            Console.WriteLine("foo");
        }
        catch 
        {
            throw;
        }
    }

but instead, returns
  public void foo()
    {
        Console.WriteLine("foo");
    }

Is broken in C# and VB

jwooley:
Fixes #587 - stripping out non-selected catch blocks. Only remove try block option if single exception left. Fixes VB and C# implementations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,14 @@
 # Code Cracker
 
-An analyzer library for C# and VB that uses [Roslyn](http://msdn.microsoft.com/en-us/vstudio/roslyn.aspx) to produce refactorings, code analysis, and other niceties.
+An analyzer library for C# and VB that uses [Roslyn](https://github.com/dotnet/roslyn) to produce refactorings, code analysis, and other niceties.
 
 Check the official project site on [code-cracker.github.io](http://code-cracker.github.io). There you will find information on how to contribute,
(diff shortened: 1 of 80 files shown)


giggio:
@jwooley You need to update your code. The current `EmptyCatchBlockCodeFixProvider` is very different, it is already using resources, for example.

giggio:
Hey, @jwooley, can you check this out? ;)

giggio:
@jwooley We can't have warning as errors disabled. What warnings are you seeing?

jwooley:
Responded to your personal email account.

jwooley:
Updated reverting the warnings as errors

PR #589, opened 2015-11-07
Reorganize StringFormatArgsAnalyzer to use the correct severities

### LINKED ISSUE #585: BUG: CC0056 (StringFormatAnalyzer) raised incorrectly
This code is incorrectly raising `CC0056`:

````csharp
string.Format ("[Line {0}:{1,2}-{3,4}:{5}]", "a", "b", "c", "d", "e", "f");
````

With message:

`The number of arguments in String.Format is incorrect.`

giggio:
Fixes #585
StringFormatArgsAnalyzer was always raising an error, when it is not an
error to have extra parameters. So we have taken an extra id, CC0111 to
use as the warning, when string.Format is being provided more arguments
than the format string is requiring.
This will now raise a warning:

    string.Format("{0}", "a", "b");

When string.Format has less arguments than the format string requires,
we are still raising CC0056, as an error, as it will generate a
runtime error.
This will now raise an error:

    string.Format("{0}{1}", "a");

Other verifications the analyzer was already checking continue as
before.

### DIFF
--- a/src/CSharp/CodeCracker/Usage/StringFormatArgsAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Usage/StringFormatArgsAnalyzer.cs
@@ -17,27 +17,27 @@ namespace CodeCracker.CSharp.Usage
         internal const string Category = SupportedCategories.Usage;
         const string Description = "The format argument in String.Format determines the number of argument, considering the {} inside. You should pass the correct number of arguments.";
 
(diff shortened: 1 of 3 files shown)


carloscds:
@giggio This code could be show a warning:
`````
var x = string.Format("{0}");
`````
What do you think about ?:

giggio:
@carloscds Yes, it should show an error, I will fix it.

giggio:
@carloscds fixed!

PR #596, opened 2015-11-17
Using and accepting 'System.GC.SuppressFinalize' (fix #590)

### LINKED ISSUE #590: BUG: DisposablesShouldCallSuppressFinalize should use full name when System is not imported
There is already a test at `DisposablesShouldCallSuppressFinalizeTests` called `AddsSystemGCWhenSystemIsNotImported` at the source, skipped. The idea is make this test pass.

This:

````csharp
                    public class MyType : System.IDisposable
                    {
                        void IDisposable.Dispose()
                        {
                            Dispose(true);
                        }
                        protected virtual void Dispose(bool disposing)
                        {
                        }
                    }
````

Should become:

````csharp
                    public class MyType : System.IDisposable
                    {
                        void IDisposable.Dispose()
                        {
                            Dispose(true);
                            System.GC.SuppressFinalize(this);
                        }
                        protected virtual void Dispose(bool disposing)
                        {
                        }
                    }
````

ElemarJR:


giggio:
@ElemarJR The analyzer change is insufficient. You are verifying the syntax, you should be verifying the symbol. This will not work with the current fix also, and will still raise a diagnostic incorrectly:

````csharp
using A = System;
public class MyType : System.IDisposable
{
    void IDisposable.Dispose()
    {
        Dispose(true);
        A.GC.SuppressFinalize(this);
    }
    protected virtual void Dispose(bool disposing)
    {

    }
}
````

Also, on the fix is incorrect, you could have a using with the scope restricted by a namespace declaration, like so:

````csharp
namespace A
{
    using System;
}
namespace B
{
    class Foo : System.IDisposable
    {
        public void Dispose()
        {
            GC.SuppressFinalize(this);
        }
    }
}
````

This will return `true` on your method that checks for usings, but it is out of scope. The way to correct this is to use the full type name, and then use the simplifier.

ElemarJR:
I got it! I will fix it. 

But, this PR improves a lot this Analyzer and Fix. I don't know if it shouldn't be accepted anyway.
I think we should use the scenarios you mentioned in another Issue.

PR #599, opened 2015-11-17
Change projects property "Treat warnings as error" to None

### LINKED ISSUE #598: Change project Erros and Warnings
Change CodeCracker projects property "Treat warnings as error" to None to avoid compiler errors.

carloscds:
Fix #598


giggio:
@carloscds instead of disabling warning as errors, we just need to disabled `AD0001` as an error. That was done on f767d631f18aea0a940da2f8370796a417c724b7. This way we are not stopped from other analyzer's errors and we still get to keep warnings away. All we needed to do is add `<WarningsNotAsErrors>AD0001</WarningsNotAsErrors>` to the .csproj and .vbproj files.

PR #602, opened 2015-11-19
Fix should get the right expression to process (fixes #597)

### LINKED ISSUE #597: BUG on CC0049
This will throw:

````csharp
if (obj is ProjectCompilation == false) return false;
````

With `ProjectCompilation` being a struct and `obj` being of type `object`.

### PR DESCRIPTION
**Pull request by ElemarJR**

### CODE REVIEW

File: test/CSharp/CodeCracker.Test/Usage/SimplifyRedundantBooleanComparisonsTests.cs [hunk 1]
```diff
@@ -102,5 +103,30 @@ public class SimplifyRedundantBooleanComparisonsTests
...
+    }
+}";
+
+            var fixtest = @" 
+struct ProjectCompilation {}
+class Foo
+{
+    public void Comp(bool obj)
+    {
+        if (!(obj is ProjectCompilation)) return false;
```
GuilhermeMatheus: Is that ok write a code which will not compile, like this void returning false?

ElemarJR: I just fixed it. Thanks a lot for your feedback.

ElemarJR:
@GuilhermeMatheus thank you, again for your feedback.

I agree with you "return !exp" would be nicer... but, this PR is just fixing a bug.

Could you, please, open an Issue recommending this improvement?

GuilhermeMatheus:
Ok. I will! @ElemarJR

PR #604, opened 2015-11-19
Fix `nameof` fix to work correctly with methods with non system types

### LINKED ISSUE #594: Suggested Nameof fix incorrectly suggests wrong type
The following code fragment 

        public ActionResult ChangePassword(ChangePasswordMessageId? message, bool? Redirect = false)
        {
            var returnUrl = "ChangePassword";

Correctly identifies a place where `nameof` should be used. However it suggests using `nameof(ChangePasswordMessageID)` instead of the correct `nameof(ChangePassword)`. If I change the type of `message` it produces the correct fix, and if I change `returnUrl` to `"ChangePasswordMessage"` then it does not suggest any fix. This leads me to believe that it's correctly identifying the issue, but incorrectly finding the suggested fix because of the matching prefix, although I haven't looked at the code so I can't be sure (I'll take a look later on if you want).

Tested with code-cracker version 1.0.0-rc2 (although I believe it also existed in earlier versions).

Here is a simple test that will not work:

````csharp
public class TypeName
{
    void Foo(TypeName d)
    {
        var bar = new [] { "Foo" };
    }
}
````

And here is the test case:

````csharp
        [Fact]
        public async Task ReplacesUsingMethodName()
        {
            const string source = @"
public class TypeName
{
    void Foo(TypeName d)
    {
        var bar = new [] { ""Foo"" };
    }
}";

            const string fixtest = @"
public class TypeName
{
    void Foo(TypeName d)
    {
        var bar = new [] { nameof(Foo) };
    }
}";
            await VerifyCSharpFixAsync(source, fixtest);
        }
````

@mirhagk is working on it.

mirhagk:
Fixes #594

### DIFF
--- a/src/CSharp/CodeCracker/Design/NameOfAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Design/NameOfAnalyzer.cs
@@ -88,7 +88,7 @@ namespace CodeCracker.CSharp.Design
                         return null;
                 }
 
-                programElementName = symbol.ToDisplayParts().LastOrDefault(AnalyzerExtensions.IsName).ToString();
+                programElementName = symbol.ToDisplayParts(NameOfSymbolDisplayForamt).LastOrDefault(AnalyzerExtensions.IsName).ToString();
             }
 
             return programElementName;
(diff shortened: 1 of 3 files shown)


giggio:
@mirhagk This will not fix the bug. Here is a simple test that will not work:

````csharp
public class TypeName
{
    void Foo(TypeName d)
    {
        var bar = new [] { "Foo" };
    }
}
````

And here is the test case:

````csharp
        [Fact]
        public async Task ReplacesUsingMethodName()
        {
            const string source = @"
public class TypeName
{
    void Foo(TypeName d)
    {
        var bar = new [] { ""Foo"" };
    }
}";

            const string fixtest = @"
public class TypeName
{
    void Foo(TypeName d)
    {
        var bar = new [] { nameof(Foo) };
    }
}";
            await VerifyCSharpFixAsync(source, fixtest);
        }
````

giggio:
@mirhagk Are you looking into this?

giggio:
@mirhagk It is ok if you don't have time to fix this, but in that case we will make it up for grabs again. Only if we don't hear from you, ok?

mirhagk:
Hey sorry, got busy the last week, I should have time to fix it over the next few days

mirhagk:
Okay I've fixed that test case as well (I had only put it on the analyzer, and not on the code fix provider)

Issue #348, opened 2015-05-18
Update all existing code fixes that are doing too much work on RegisterCodeFixesAsync (VB)


giggio:
Right now most of the code fix providers are doing more work than they should on the `RegisterCodeFixesAsync` method.
They should delegate most, if not all, of the work to the code actions.

See the expected way to do this on:
https://github.com/code-cracker/code-cracker/blob/master/src/CSharp/CodeCracker/Performance/UseStaticRegexIsMatchCodeFixProvider.cs#L22

And how the correction was applied on 34c9ca4a819d180.

This is for VB only. For C# view issue #347.

jwooley:
If this is still open, I can take it.

giggio:
@jwooley go for it! Fix it on `master` so we can get it ready for v1 RTM!

giggio:
Closed on #605.

PR #605, opened 2015-11-20
#348 Update VB code fixes to reduce work done in register code fix method

jwooley:
resolves issue #348


carloscds:
@jwooley Great change Jim! Thanks!

Issue #626, opened 2015-12-15
CC0003 is shown even for catches which end with throw


Duetting:
Imho CC0003 shouldn't be shown for catch blocks which end with throw:
```csharp
                var disposable = new DisposableObject();
                try
                {
                    ...
                }
                catch
                {
                    disposable.Dispose();
                    throw;
                }
```

giggio:
Agreed!

giggio:
Closed on #631

PR #631, opened 2015-12-18
Fix for #626 CC0003 is shown even for catches which end with throw

thorgeirk11:
This PR fixes the #626 issue.


carloscds:
@thorgeirk11 Good job!

PR #633, opened 2015-12-24
Fixes #625 CC0001 is shown for variables with type dynamic and object

### LINKED ISSUE #625: CC0001 is shown for variables with type dynamic and object
`CC0001` should not be reported for dynamic like this:

````csharp
dynamic foo = new object();
````

See discussion bellow. It only happens when the right hand side of the assignment is an object. This already does not report `CC0001`:

````csharp
dynamic foo = 1;
````

Original comment:

> Imho it doesn't make sense and can generate errors if the type definition of a variable is changed from dynamic to var. Especially when using COM we use variables of type dynamic. The type dynamic should never be replaced with var, as it's intentionally used.

thorgeirk11:


thorgeirk11:
Sorry Incorrect branch

PR #651, opened 2016-01-08
Fix #650 by rearranging the order the tree is modified 

### LINKED ISSUE #650: With more than one constructor, IntroduceFieldFromConstructorCodeFixProvider can replace wrong constructor
When the `IntroduceFieldFromConstructorCodeFixProvider` is run on a class with more than one constructors, the first constructor is always replaced.

The bug is caused by this line in `IntroduceFieldFromConstructorCodeFixProvider.cs`:
```
            newClass = newClass.ReplaceNode(newClass.DescendantNodes().OfType<ConstructorDeclarationSyntax>().First(), newConstructor);
```

I have a pull request coming shortly...

For Example
----------------
This:
```
 using System;

    namespace ConsoleApplication1
    {
        class TypeName
        {
            public TypeName()
            {
            }

            public TypeName(int par)
            {
            }
        }
    }
```
Should be:
```
    using System;

    namespace ConsoleApplication1
    {
        class TypeName
        {
            private readonly int par;

            public TypeName()
            {
            }

            public TypeName(int par)
            {
               this.par = par;
            }
        }
    }
```
But actually:
```
    using System;

    namespace ConsoleApplication1
    {
        class TypeName
        {
            private readonly int par;

            public TypeName(int par)
            {
               this.par = par;
            }

            public TypeName(int par)
            {
            }
        }
    }
```

haroldhues:
Refactored version swaps the constructor and then adds a member if needed

### DIFF
--- a/src/CSharp/CodeCracker/Refactoring/IntroduceFieldFromConstructorCodeFixProvider.cs
+++ b/src/CSharp/CodeCracker/Refactoring/IntroduceFieldFromConstructorCodeFixProvider.cs
@@ -13,7 +13,7 @@ using System.Threading.Tasks;
 
 namespace CodeCracker.CSharp.Refactoring
 {
-    [ExportCodeFixProvider(LanguageNames.CSharp, Name =nameof(IntroduceFieldFromConstructorCodeFixProvider)), Shared]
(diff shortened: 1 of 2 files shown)


giggio:
Tks!

PR #652, opened 2016-01-08
Fix #649 by looking up the parameter as a named argument.

### LINKED ISSUE #649: BUG on CC0057 (unused parameter) with named parameters
This will be fixed incorrectly:

````vb
Sub Main()
    Foo(c:=3, b:=3)
End Sub
Sub Foo(b As Integer, Optional c As Integer = 1)
    c = 2
End Sub
````

To this:

````vb
Sub Main()
    Foo(b:=3)
End Sub
Sub Foo(Optional c As Integer = 1)
    c = 2
End Sub
````

It should had removed the `b` parameter on the `Foo` invocation, but it removed `c`. It is removing by position, not by reference. The correct code after the fix should be:

````vb
Sub Main()
    Foo(c:=3)
End Sub
Sub Foo(Optional c As Integer = 1)
    c = 2
End Sub
````

Problem also happens with C#:

This:

````csharp
void Main()
{
    Foo(c: 3, b: 3);
}
void Foo(int b, int c = 1)
{
    b = 2;
}
````

Is incorrectly fixed to:

````csharp
void Main()
{
    Foo(c: 3);
}
void Foo(int b)
{
    b = 2;
}
````

It should be instead:

````csharp
void Main()
{
    Foo(b: 3);
}
void Foo(int b)
{
    b = 2;
}
````

### PR DESCRIPTION
**Pull request by adraut**

### CODE REVIEW

File: src/CSharp/CodeCracker/Usage/UnusedParametersCodeFixProvider.cs [hunk 1]
```diff
@@ -49,10 +49,10 @@ public sealed override Task RegisterCodeFixesAsync(CodeFixContext context)
             if (parameterList.Parameters.First().ToString().Contains("this")) parameterPosition--;
             var newParameterList = parameterList.WithParameters(parameterList.Parameters.Remove(parameter));
             var foundDocument = false;
-            var semanticModel = await document.GetSemanticModelAsync(cancellationToken);
+            var semanticModel = await document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
```
adraut: Moving up to potentially save a context switch.

src/VisualBasic/CodeCracker/Usage/UnusedParametersCodeFixProvider.vb
```diff
@@ -45,10 +45,10 @@ Namespace Usage
             Dim parameterPosition = parameterList.Parameters.IndexOf(parameter)
             Dim newParameterList = parameterList.WithParameters(parameterList.Parameters.Remove(parameter))
             Dim foundDocument = False
-            Dim semanticModel = Await document.GetSemanticModelAsync(cancellationToken)
+            Dim semanticModel = Await document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(False)
```
adraut: Moving up to potentially save a context switch.

src/VisualBasic/CodeCracker/Usage/UnusedParametersCodeFixProvider.vb
```diff
@@ -71,9 +71,15 @@ Namespace Usage
                     Dim arguments = If(objectCreation IsNot Nothing,
                         objectCreation.ArgumentList,
                         methodIdentifier.FirstAncestorOfType(Of InvocationExpressionSyntax).ArgumentList)
-                    If parameter.Modifiers.Any(Function(m) m.IsKind(SyntaxKind.ParamArrayKeyword)) Then
+
+                    ' Attempt to find the parameter as a named argument.  Named arguments can only appear once in the argument list.
+                    Dim namedArg = arguments.Arguments.Where(Function(arg) arg.IsNamed).OfType(Of SimpleArgumentSyntax).SingleOrDefault(Function(arg) arg.NameColonEquals.Name.Identifier.Text = parameter.Identifier.Identifier.Text)
```
adraut: I'm not sure if this is the most efficient way to do this.  Other ideas are appreciated.

src/CSharp/CodeCracker/Usage/UnusedParametersCodeFixProvider.cs
giggio: We have been just leaving `.ConfigureAwait(false)` on every Task.

src/VisualBasic/CodeCracker/Usage/UnusedParametersCodeFixProvider.vb
giggio: I have been using `IsKind` instead of `OfType`, as it is much faster to check a value than to reflect on the type.

Issue #645, opened 2016-01-05
CC0016 should not copy the variable with readonly members


Haacked:
Suppose I have the following class:

```csharp
public class HttpHandlerRouteHandler<THttpHandler> : IRouteHandler where THttpHandler : IHttpHandler
{
    readonly Func<RequestContext, THttpHandler> _handlerFactory;

    public HttpHandlerRouteHandler(Func<RequestContext, THttpHandler> handlerFactory)
    {
        if (handlerFactory == null) throw new ArgumentNullException(nameof(handlerFactory));
        _handlerFactory = handlerFactory;
    }

    public IHttpHandler GetHttpHandler(RequestContext requestContext)
    {
        return _handlerFactory(requestContext);
    }
}
```

Note that `_handlerFactory` is a `readonly` field so it can't generally be changed on more than one thread. In this particular case, it also can't be null. So there is no need to copy the event handler to a variable before firing it.

But Code Cracker still recommends I copy it. Not a big deal since I'm betting this is tricky to get right. But thought I'd mention it. :smile:

Summarizing:

If the variable **is readonly** and it is being assigned on constructor or directly on the field declaration, do not raise `CC0016` anymore.

We will create another issue to raise an error when a readonly reference field that is never assigned is used for anything.

@Vossekop is working on it.

carloscds:
@Haacked Thanks for your comment about this!

giggio:
@Haacked On the `GetHttpHandler` method there are only two options for `_handlerFactory`: null or not null, and, as you said, that cannot be changed. Not a single thread, but at all, as being readonly only allows it's value to change during object construction. So, there is no way that the value is going to change, and therefore, there is no need to check.

A better approach would be to simply `_handlerFactory?.Invoke`:

````csharp
public IHttpHandler GetHttpHandler(RequestContext requestContext)
{
    return _handlerFactory?.Invoke(requestContext);
}
````

But that would fail unless we declare that `THttpHandler` is a `class`:

````
public class HttpHandlerRouteHandler<THttpHandler> : IRouteHandler where THttpHandler : class, IHttpHandler
````

So, yes, the copying is unnecessary, but the check for null is not. And we cannot reliably fall back to `?.`.

So, I am suggesting that we simply remove the copy when the field is readonly.

Also, there is another bug to fix. I applied the fix, and it removed the return statement, ending up like this:

````csharp
public IHttpHandler GetHttpHandler(RequestContext requestContext)
{
[...]

PR #660, opened 2016-01-14
#645 Better handling of CC0016 in situations where it concerns a readonly …

Vossekop:
…variable.

Also a whole lot of new unit tests for CC0016 concerning these readonly scenarios.

Wow, something went horribly wrong with my previous pull request. Let me try to not delete the branch prematurely this time...

### DIFF
--- a/src/CSharp/CodeCracker/Design/CopyEventToVariableBeforeFireAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Design/CopyEventToVariableBeforeFireAnalyzer.cs
@@ -3,7 +3,9 @@ using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using System;
+using System.Linq;
 using System.Collections.Immutable;
+using System.Collections.Generic;
 
 namespace CodeCracker.CSharp.Design
 {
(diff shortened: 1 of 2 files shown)


giggio:
Good job, and very nice way to solve the problem. See 82fba2a for minor improvements, if you want to discuss what I did let me know.

Vossekop:
I can see the reason behind most changes. Using the IsKind method seems arbitrary in most situations, is this the preferred way according to the Roslyn team, or is it maybe faster than the 'is' operator? The original code had all 'is' operators so I followed that where possible. Also I'll try to refrain from my braces rich code style next time, it is an acquired taste, I know ;-)

giggio:
`IsKind` is much faster than using any type of Reflection. It is a simple property retrieval against all the weight of Reflection. And you are right, we had been using it, it was how Microsoft showed their samples.
And don't worry about it, this is the process of Open Source, we improve each other's code. :)

Vossekop:
Not because I didn't believe you, but because I wanted to know what the difference was I did some performance tests. The results were interesting to say the least. I performed 2 tests (all in debug mode btw with 100 000 000 iterations):

(symbol.Kind == SymbolKind.Field) vs (symbol is IFieldSymbol) 
and
(statement.IsKind(SyntaxKind.Block) vs (statement is BlockStatement)

In the first test, the property check is the clear winner, just 25% of the time it takes to run the 'is' operator.
However in the second test, the method call is 10 times slower than the 'is' operator. Ofcourse the debugger stacks the deck in favour of the 'is' operator by not optimizing calls to the method, but still the 'is' operator can hardly be called slow (300ms for 100 000 000 operations). This is probably because 'is' is not just any type of reflection, it gets mapped directly to an IL operation (isinst to be precise) see more info here: http://stackoverflow.com/questions/1134004/c-sharp-is-operator-is-that-reflection

I'll try to whip up a simple console app for an easy repeatable perf test.

Vossekop:
Ok, so after some more testing I can definitely say the IsKind method is about as fast as using the simple property check and the is operator with an interface is about 5 times slower. But using the is operator with a specific class is 3 times faster and with a base class it is even 10 times faster.

giggio:
Interesting findings. If you can push a repo so we can check that would be awesome.

giggio:
Ok, I checked. See repo: https://github.com/giggio/roslynperf
Here are my findings:

````
IsKind Match: 42.21
IsKind No Match: 42.50
RawKind.Equals: 44.79
is Class: 29.53
is Struct (no match): 30.72
is interface (no match): 40.39
as Class: 25.92
as Struct (no match): 26.47
as interface (no match): 39.91
if then direct cast: 60.57
````

You can read the tests here:

https://github.com/giggio/roslynperf/blob/master/RoslynPerf/Program.cs

My conclusions are:

`IsKind` is slower when compared to `is` with classes, and about the same with interfaces.
It is not worth to check with `IsKind` then convert with `()`. It is much slower than using `as` (from 50% to 140%).

@code-cracker/owners check this findings.

With this results in hand, I am thinking we should never use `IsKind` or `Kind`, or `RawKind` if we have an equivalent class.

Vossekop:
Sorry for the somewhat late reaction, got a new internet connection that wouldn't play ball the last 3 days. I also found that there was a difference in speed if the as or is operator was used for the actual class or for a base class (the latter being faster than the former).

Issue #658, opened 2016-01-14
BUG on CC0047 (PropertyPrivateSet) suggests to make a property set private when the property is already private


giggio:
This should not raise `CC0047`:

````csharp
class Foo
{
    private int P { get; set; }
}
````

Vossekop:
I'll pick this one up as well!

giggio:
@Vossekop go for it.

giggio:
Closed on #658.

PR #661, opened 2016-01-14
#658 Added check to PropertyPrivateSet so it doesn't show when the property is already private.

Vossekop:
…y is already private.


giggio:
@Vossekop There was a message problem on the tests, I fixed it on 5c439ac.

Vossekop:
Oops, my bad, I'll keep an eye on that next time.

PR #666, opened 2016-01-19
Fixes #665 by checking if any Parent is using or return

### LINKED ISSUE #665: CC0022 when theer is ternary operator in using
Following code generates CC0022
```C#
using (color.HasValue ? new ConsoleColorContext(color.Value) : null)
```
and when I click on "Show potential fixes" DisposableVariableNotDisposedCodeFixProvider crashes

martinzimmermann:


carloscds:
Hi @martinzimmermann What issue are you fixing ?

carloscds:
Hi @martinzimmermann. I just test with this code:
`````CSharp
 public class CSharpClass
    {
        struct ConsoleColorContext : IDisposable
        {
            public void Dispose()
            {
            }
        }

        void Foo()
        {
            ConsoleColorContext? color = null;
            var x = color.HasValue ? new ConsoleColorContext(color.Value) : null;
        }
    }

`````
And I already hae a crash. Can you check ?

martinzimmermann:
I can verify that this still produces a crash, I will have a look into it.

martinzimmermann:
Hi @carloscds,
This should fix it now. 

Maybe there are other expressions that do not interfere with the evaluation of if an object creation should raise the Diagnostic or not. Currently I identified that Parentheses, Casts and Ternary Operators do not interfere with the evaluation.
If you can think of anything else, just let me know and I will fix this too.

carloscds:
@martinzimmermann Good job!

Issue #508, opened 2015-09-13
Convert method to property


giggio:
This:

````csharp
class Foo
{
    int Bar() => 1;
    int Baz()
    {
        return 2;
    }
    void Using()
    {
        var v = Bar() + Baz();
    }
}
````

Becomes:

````csharp
class Foo
{
    int Bar => 1;
    int Baz
    {
        get
        {
            return 2;
        }
    }
    void Using()
    {
        var v = Bar + Baz;
    }
}
````

Can only be offered if method does not return void. Method invocations have to be changed to property references.

Category: Refactoring
Id: CC0103
Severity: Hidden

Shiney:
I'd like to start on this task.

Do we need to stop the code fix from being applicable if the method is overloaded?

carloscds:
@Shiney I think not because you won't change method signature.

Shiney:
But you can't have a method and a property with the same name.

carloscds:
@Shiney Can you put a sample of this ?

giggio:
@Shiney I added you to the collaborators team so we can assign this issue to you. You may start working on it.

Answering your question, I can't think of a way to solve the overloaded method problem. Simply ignore the method if it has an overload.

giggio:
@Shiney Also, remember to fork from the vnext branch, and to request to pull into it. Not from master, which is stabilizing now for 1.0.

Shiney:
class A{
public int Foo(int a){ return a;}

public int Foo{get{return 314;}}

}

Does not compile.


Am I supposed to look up and down the whole inheritance tree to find out if this method was overloaded everywhere? Or would it be simpler and less confusing for the user to just get sometimes non-compilable code out of the code fix?

giggio:
@Shiney A first quick check would be to ask for the type symbol, check all members with the same name, if there is more than one, return. You should use symbolic analysis as you may be dealing with a partial class. Then, if that succeeds, you should ask for all parents and children in the class hierarchy if and see if you find a member with the same name. Please put that as on the common project as an extension, as it might be useful for other analyzers/fixers.

Shiney:
I'm a bit confused about how to replace the old references. Is there a good example of looking at references and changing them?

giggio:
On the code fix you can use the `SymbolFinder` class. There are examples on the CodeCracker source you can use.

giggio:
Already present in latest VS/Roslyn. Feature dropped.

PR #668, opened 2016-01-23
Convert method to property

Shiney:
Created analyzer and code fix provider for convert to property.

It doesn't support virtual or abstract methods or interfaces. #508

### DIFF
--- a/src/CSharp/CodeCracker/CodeCracker.csproj
+++ b/src/CSharp/CodeCracker/CodeCracker.csproj
@@ -79,7 +79,9 @@
     <Compile Include="Refactoring\AllowMembersOrderingCodeFixProvider.StyleCop.cs" />
     <Compile Include="Refactoring\AllowMembersOrderingCodeFixProvider.Base.cs" />
     <Compile Include="Refactoring\ChangeAsToCastAnalyzer.cs" />
+    <Compile Include="Refactoring\ConvertMethodToPropertyAnalyzer.cs" />
     <Compile Include="Refactoring\ChangeAsToCastCodeFixProvider.cs" />
+    <Compile Include="Refactoring\ConvertMethodToPropertyCodeFixProvider.cs" />
     <Compile Include="Refactoring\ExtractClassToFileAnalyzer.cs" />
     <Compile Include="Refactoring\MakeMethodNonAsyncCodeFixProvider.cs" />
     <Compile Include="Refactoring\MergeNestedIfAnalyzer.cs" />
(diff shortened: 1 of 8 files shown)


giggio:
@Shiney Please update the analyzer and code fix to use localizable strings for title, etc. See the `nameofanalyzer` and fix for an example.

giggio:
@Shiney This is raising a diagnostic, and it shouldn't. If the code fix is applied it will create a compiler error:

````
class C
{
    public int I(string i) => 1;
    public int I() => 1;
}
````

**Comment by Shiney** [comment]
@giggio if I put that code into a test and call 
```CSharp
            await VerifyCSharpHasNoDiagnosticsAsync(test);
```

Then the test passes. Do you have any idea what I've done wrong in the test?

giggio:
I have no idea why. Have you tried running? I get the problem there, see:

error

giggio:
I noticed that this is not coming from us. Maybe from VS/Roslyn directly?
So, anyway, disregard, I will check the rest.

giggio:
@Shiney I have found a problem when the class is implementing an interface or inheriting from a base class. It's methods are being changed without change the base class or interface.

So, here, changing `I` to become a property on `Foo` should also change the interface and `Foo2`. Same thing for the `Virtual` and `Abstract` methods.

````csharp
abstract class BaseFoo
{
    public abstract int Abstract();
    public virtual int Virtual()
    {
    }
}
interface IFoo
{
    int I();
}
class Foo : IFoo
{
    public int I()
    {
    }
}
class Foo2 : BaseFoo, IFoo
{
    public override int Abstract()
    {
    }
    public override int Virtual()
    {
    }
    public int I()
    {
    }
}
````

I noticed that Microsoft is also suggesting this fix, but theirs is broken, it suggests to change even when there is an overload. I think they released it with Roslyn 1.1, as I don't think it was there before. I was going to suggest we stop this feature, but since theirs is broken, let's make ours better. :)
Also, as you have to update those items above, I think you should also support virtual or abstract methods, and also interface methods.

Shiney:
Do you think the changing even if there is an overload is bad? It seems to me like that is a less confusing experience for the user as they won't start getting confused when they can't make the change they want to make.

giggio:
We should not offer a change if there is an overload, as that would break the code.

Shiney:
I think we should probably just drop the feature it gets really complicated really fast especially since there are so many different ways of implementing a method. For example if a method with exactly the same signature is a member of two interfaces and you do the convert method to property on one of the interfaces what is supposed to happen when you do the convert method to property? I'm not convinced there's a good story about what the right thing to do is. If theres an out of the box solution it's probably bettter if they just use that.

giggio:
@Shiney yes, there is an out of box solution for that, made by the roslyn team. What do you think, @code-cracker/owners, should we drop it?

carloscds:
@giggio I think that already exsits, we shouldn't develop again!

giggio:
Ok, so feature dropped.

Issue #307, opened 2015-04-07
Add property to constructor


giggio:
This:

````csharp
class Foo
{
    public int Bar { get; set; }
}
````

Becomes:

````csharp
class Foo
{
    public Foo(int bar)
    {
        this.Bar = bar;
    }
    public int Bar { get; set; }
}
````

If the constructor already exists, add to it:

This:

````csharp
class Foo
{
    public Foo()
    {
    }
    public int Bar { get; set; }
}
````

Becomes:

````csharp
class Foo
{
    public Foo(int bar)
    {
        this.Bar = bar;
    }
    public int Bar { get; set; }
}
````

If more than one constructor exists, offer one fix for each constructor, and another for all constructors.

Diagnostic Id: `CC0083`
Severity: Hidden
Category: Refactoring

carloscds:
@giggio How I know what property can put in constructor ?

Look:
````csharp

class Foo
{
    public int Bar { get; set; }
    public int Some {get; set;}
}
````

jwooley:
I can see this becoming an anti-pattern quickly. What happens if I have 20 properties. I would hate to have a constructor for each property even if they are optional. For me a constructor with that many properties should be refactored. This makes more sense for setting private fields or properties without public setters.

giggio:
@carloscds we would have a diagnostic for each property that isn't already being set on the constructor. On your example, we would have 3 diagnostics:

* Add Bar assignment to constructor
* Add Some to constructor
* Add all properties assignments to constructor

giggio:
@jwooley We would not generate 20 constructor, only one.
We could expand this issue to include private fields. Properties without public setters should already be included.

jwooley:
I was referring to 20 parameters to the single constructor.






Sent from Windows Mail





From: Giovanni Bassi
Sent: ‎Wednesday‎, ‎April‎ ‎8‎, ‎2015 ‎3‎:‎27‎ ‎PM
To: 'code-cracker/code-cracker'
Cc: Jim Wooley





@jwooley We would not generate 20 constructor, only one.
We could expand this issue to include private fields. Properties without public setters should already be included.

carloscds:
@giggio I guess that this issue need more discussion because it's not totally clear.

giggio:
[...]

PR #674, opened 2016-01-26
add property to constructor #307

julianosaless:
New feature #307


carloscds:
@julianosaless When I have many properties and choose Document option at fixer, only one property is included in constructor. You need to implement FixAllProvider, to correct all properties at same time, if user needs.

julianosaless:
@carloscds, I added the FixAllProvider :)

carloscds:
@julianosaless Can you check your tests, because I have a decrease um code coverage,

julianosaless:
@carlos  i think that probleam will be resolved with this new  comit

carloscds:
Good job @julianosaless

julianosaless:
thank you so much @carloscds.

Issue #680, opened 2016-01-27
BUG: CC0091 (make static) changes references incorrectly when used as a method group and in conjunction with another method invocation


jakubsuchybio:
This is producing an incorrect fix:

````csharp
class Bar
{
    void ShouldBeStatic()
    {
    }
    void Caller()
    {
        Foo.M(new Baz(ShouldBeStatic));
    }
}
class Foo
{
    public static void M(Baz b) { }
}
class Baz
{
    public Baz(Action a)
    {
    }
}
````

This is the current fix, broken (other classes omitted):

````csharp
class Bar
{
    static void ShouldBeStatic()
    {
    }
    void Caller()
    {
        M(new Baz(ShouldBeStatic));
    }
}
````

This is the expected fix (other classes omitted, only `static` added to method, nothing else changed):

````csharp
class Bar
{
    static void ShouldBeStatic()
    {
    }
    void Caller()
    {
        Foo.M(new Baz(ShouldBeStatic));
    }
}
````

Here is the original description from @jakubsuchybio:

> I found another error in CC0091 which somehow changes another method.
> Dunno why it is changing things in another method not related to the changed method.
> ```c#
>         private void ResetButton_Confirm(Gump sender, bool response)
>         {
>             if( response )
>             {
>                 PerConfig.DefaultMacros();
>                 MacroProvider.InstanceIndex = 0;
>             }
> 
>             GMacroEditorForm.Open();
>         }
> 
>         private void ResetButton_OnClick(Gump g)
>         {
>             this.Close();
>             Gumps.Desktop.Children.Add( new GMessageBoxYesNo(
>                 ClientLocalization.GetString( "#319#Smazat všechna\rvlastní makra?" ),
>                 true,
>                 ResetButton_Confirm ) );
>         }
> ```
> Translates into:
> ```c#
>         private static void ResetButton_Confirm(Gump sender, bool response)
>         {
>             if( response )
>             {
>                 PerConfig.DefaultMacros();
>                 MacroProvider.InstanceIndex = 0;
>             }
> 
>             GMacroEditorForm.Open();
>         }
> 
>         private void ResetButton_OnClick(Gump g)
>         {
>             this.Close();
>             Add( new GMessageBoxYesNo(
>                             ClientLocalization.GetString( "#319#Smazat všechna\rvlastní makra?" ),
>                             true,
>                             ResetButton_Confirm ) );
>         }
> ```
> Here is a picture of preview:
[...]

Issue #677, opened 2016-01-26
CC0091 on method Foo does not replace occurences like this.Foo used in delegates


jakubsuchybio:
Hi.
I created code as proof of bug:
```
		class X
		{
			public static int f = 0;

			public void Foo()
			{
				f = 5;
			}
		}

		class Y : X
		{
			public event Action Changed;

			public void Goo()
			{
				Changed += this.Foo;
			}
		}
```
When I apply correction of CC0091 onto Foo, occurrence of this.Foo; doesn't change and then this occurrence gives error, because it doesn't know this.Foo only Foo.
Regards

Edit: And I am stupid, because this issue has already been reported. Sorry

giggio:
Bug confirmed on the code fix. It should change to:

````csharp
class X
{
    public static int f = 0;

    public static void Foo()
    {
        f = 5;
    }
}

class Y : X
{
    public event Action Changed;

    public void Goo()
    {
        Changed += Foo;
    }
}
````

It should had changed `this.Foo` to `Foo` on the `Goo` method body.

martinzimmermann:
I would like to fix this

giggio:
@martinzimmermann go for it.

giggio:
Closed on #691.

PR #688, opened 2016-01-28
Bug fix cc0091 not working when used with delegates

martinzimmermann:
When a Method, that should be made static is used somewhere like `new  X().MethodToBecomeStatic();`. I fix it to `new X(); X.MethodToBecomeStatic()` to ensure that the program logic is not broken.

### DIFF
--- a/src/CSharp/CodeCracker/Design/MakeMethodStaticCodeFixProvider.cs
+++ b/src/CSharp/CodeCracker/Design/MakeMethodStaticCodeFixProvider.cs
@@ -72,15 +72,15 @@ namespace CodeCracker.CSharp.Design
             var unchangedRoot = root;
             var memberAccess = root.GetCurrentNode(diagnosticNode).FirstAncestorOrSelfOfType<MemberAccessExpressionSyntax>();
             if (memberAccess == null) return unchangedRoot;
-            if (IsTypeOfExpresion<ThisExpressionSyntax>(memberAccess.Expression))
+            if (IsTypeOfExpression<ThisExpressionSyntax>(memberAccess.Expression))
             {
                 var newMemberAccessParent = memberAccess.Parent.ReplaceNode(memberAccess, memberAccess.Name)
                     .WithAdditionalAnnotations(Formatter.Annotation);
                 return root.ReplaceNode(memberAccess.Parent, newMemberAccessParent);
             }
-            else if (IsTypeOfExpresion<ObjectCreationExpressionSyntax>(memberAccess.Expression))
+            else if (IsTypeOfExpression<ObjectCreationExpressionSyntax>(memberAccess.Expression))
             {
-                var newObjectCreationExpression = SyntaxFactory.ExpressionStatement(GetObjectCreationExpresion(memberAccess.Expression));
+                var newObjectCreationExpression = SyntaxFactory.ExpressionStatement(GetObjectCreationExpression(memberAccess.Expression));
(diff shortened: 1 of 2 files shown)


jwooley:
If it is a static method, it should not be called on the type, not the instance. It should be changed to 

T.MethodToBecomeStatic() where T is the type of X.

martinzimmermann:
Yes, this is done, but the constructor call is preserved. 

What I mean with this code:
```csharp
using System;
namespace ConsoleApplication1
{
    class Type1
    {
        public static bool Initialized = false;
        private int i;
        public Type1()
        {
            i = 0;
            Initialized = true;
        }
        public bool Foo()
        {
            return Initialized;
        }
    }
    class Type2
    {
        private int i = 0;
        public void Bar() {
            i++;
            var a = new Type1().Foo();  // <- original, should be refactored
            Console.WriteLine(a ? ""Initialized"" : ""Not Initialized""); //  <- this prints "Initialized"
        }
    }
}
```
my fix would add a `new Type1();` as a new statement and replace the original `new Type1()` with `Type1` so the code refactors to this:
```csharp
using System;
namespace ConsoleApplication1
{
    class Type1
    {
        public static bool Initialized = false;
        private int i;
        public Type1()
        {
            i = 0;
            Initialized = true;
        }
        public static bool Foo()
        {
            return Initialized;
        }
    }
    class Type2
    {
        private int i = 0;
        public void Bar() {
            i++;
            new Type1();  // <- this is added
            var a = Type1.Foo();  // <- new Type1() is replaced with Type1
            Console.WriteLine(a ? "Initialized" : "Not Initialized");  //  <- this prints "Initialized"
        }
    }
}
```
instead of just replacing the `new Type1()` with `Type1 `:
```csharp
using System;
namespace ConsoleApplication1
{
    class Type1
    {
        public static bool Initialized = false;
        private int i;
        public Type1()
        {
            i = 0;
            Initialized = true;
        }
        public static bool Foo()
        {
            return Initialized;
        }
    }
    class Type2
    {
        private int i = 0;
        public void Bar() {
            i++;
            var a = Type1.Foo();  // <- only new Type1() is replaced with Type1
            Console.WriteLine(a ? "Initialized" : "Not Initialized");  //  <- this prints "Not Initialized"
        }
    }
}
```
In this example you can see that the original code would print "Initialized". 
The first fix, with the `new Type1()`; inserted would also print "Initialized".
The second fix without the `new Type1();` would print "Not Initialized".

I think that fixes shouldn't change the behavior of the code, so I would prefer the first fix.
But if you want to have the second fix I can also change the code so that no constructor call is added.

giggio:
PR #691 is fixing this, along with several other problems on `CC0091`. I was working on it from issue #680. It conflict with this one.
@martinzimmermann I also noticed you commented on #680. When you want to start working on an issue, please check if there isn't somebody already working on it. The first one assigned will have priority. This is explaining on the contributing doc, so we don't have 2 or more people working on the same issue.
Also, I am not sure we should keep the constructor call. #691 is not keeping it. I understand the collateral effects of a constructor call, but there shouldn't be any, besides the object construction itself, it is bad design. I believe that the most common use case would be to not keep it. But I understand some people might want to keep it.
We could offer to code actions for this scenario. But maybe this is too much. @code-cracker/owners what do you think?

martinzimmermann:
@giggio: I’m very sorry for the inconvenience. I was initially fixing #677 this fix also accidentally has fixed #680, much like PR #691 also fixes #677.
Anyway thanks for the feedback, next time I will be more careful.

Issue #684, opened 2016-01-27
CC0006 should be checking if for-looped array is enumerable by foreach


jakubsuchybio:
Hi.
CC0006 should be checking if for-looped array is enumerable by foreach, if it is not, the CC0006 provider shouldn't be applied to that for-loop.
For example I have this loop:
```c#
var list = new RectangleList();
for( int m = 0; m < list.Count; m++ )
{
    var rectangle5 = list[m];
    if( Renderer.SetViewport( x + rectangle5.X, y + rectangle5.Y, rectangle5.Width, rectangle5.Height ) )
    {
        gump.Render( x, y );
    }
}
```
Where RectangleList is custom indexer which doesn't implement Enumerator
```c#
	internal class RectangleList
	{
		private int m_Count;
		private Rectangle[] m_Rects = new Rectangle[8];

		public int Count
		{
			get
			{
				return this.m_Count;
			}
		}

		public Rectangle this[int index]
		{
			get
			{
				if( ( index < 0 ) || ( index >= this.m_Count ) )
				{
					throw new IndexOutOfRangeException();
				}
				return this.m_Rects[index];
			}
		}
...
```
In this instance, CC0006 is showed on the above for-loop, but it shouldn't, because, as I said, RectangleList doesn't implement enumerator and is not enumerable by foreach.

giggio:
Bug confirmed. Here is a simplified test case:

````csharp
class Foo
{
    void Bar()
    {
        var list = new MyList();
        for (int i = 0; i < list.Count; i++)
        {
            var item = list[i];
        }
    }
}
class MyList
{
    public int Count => 1;
    public int this[int index] => 1;
}
````

PR #694, opened 2016-01-30
Fix change to for when there was an assignment and verify for Enumerable

### LINKED ISSUE #683: BUG: CC0006 wrongly tries to convert for into foreach when there would be an assignment to the iteration variable
Hi.
yes there is a pattern of creating local variable from indexed array, but after it there is write operation into indexed array, which can't be changed into foreach which is read-only loop. I think.
```c#
					for( int i = 0; i < buffer.Length; i++ )
					{
						var temp = buffer[i];
						buffer[i] = Soucet( buffer[i], lastByte );
						lastByte = temp;
					}
```
Translates via CC0006 into:
```c#
					foreach( var temp in buffer )
					{
						temp = Soucet( temp, lastByte );
						lastByte = temp;
					}
```
Which doesn't make sense, because I can't write into temp.

giggio:
Closes #683 and #684


giggio:
I will do #684 together with this one.

PR #697, opened 2016-02-01
Fix explicit property when making autoprop

### LINKED ISSUE #687: CC0017 wrong change with explicitly implemented properties, but can be fixed
Hi.
I came across some code, that I didn't even know was possible, but there it gives wrong output after fix with CC0017. See:
```c#
		interface ITranslucent
		{
			float Alpha { get; set; }
		}

		class Foo : ITranslucent
		{
			private float _alpha;
			float ITranslucent.Alpha
			{
				get
				{
					return _alpha;
				}

				set
				{
					_alpha = value;
				}
			}

			public Foo()
			{
				_alpha = 0.0f;
			}
		}
```
Translates into:
```c#
		interface ITranslucent
		{
			float Alpha { get; set; }
		}

		class Foo : ITranslucent
		{
			float ITranslucent.Alpha { get; set; }

			public Foo()
			{
				Alpha = 0.0f;
			}
		}
```
And that gives me error in constructor, that Alpha doesn't exist in current context.

But when this Interface isn't implemented it gives me 2 options:
image
and
image

When the property is not implemented explicitly like this:
```c#
		interface ITranslucent
		{
			float Alpha { get; set; }
		}

		class Foo : ITranslucent
		{
			private float _alpha;
			public float Alpha
			{
				get
				{
					return _alpha;
				}

				set
				{
					_alpha = value;
				}
			}

			public Foo()
			{
			}
		}
```
It translates into:
```c#
		interface ITranslucent
		{
			float Alpha { get; set; }
		}

		class Foo : ITranslucent
		{
			public float Alpha { get; set; }

			public Foo()
			{
			}
		}
```
Which is correct.

So I wound suggest to implement into CC0017 to change Explicit properties to public ones, which should be equivalent think.

giggio:
Closes #687

### DIFF
--- a/src/CSharp/CodeCracker/Extensions/CSharpAnalyzerExtensions.cs
+++ b/src/CSharp/CodeCracker/Extensions/CSharpAnalyzerExtensions.cs
@@ -459,6 +459,9 @@ namespace CodeCracker
             return null;
         }
 
+        public static TNode FirstAncestorOfKind<TNode>(this SyntaxNode node, params SyntaxKind[] kinds) where TNode : SyntaxNode =>
+            (TNode)FirstAncestorOfKind(node, kinds);
+
         public static IEnumerable<TNode> OfKind<TNode>(this IEnumerable<SyntaxNode> nodes, SyntaxKind kind) where TNode : SyntaxNode
         {
             foreach (var node in nodes)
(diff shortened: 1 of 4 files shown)


carloscds:
@giggio This branch has conflict.

giggio:
@carloscds fixed

PR #704, opened 2016-02-01
Fix nested interpolations

### LINKED ISSUE #690: Bug: CC0048 nested interpolations doesn't fix in one iteration
Hi.
When I have nested string.Format , then applying CC0048 will only solve first or selected one of them.
For example on this code:
```c#
var foo = string.Format( "{0}", string.Format( "{0}", 1 ) );
```
It will fix only one of them in one go. I think it would be good to solve them all at once.

Edit: Sorry I forgot to mention by using FixAll in Solution it will only fix one of them instead of both. 
When those string.Formats are concatenated they are fixed correctly. Like in here:
```c#
var foo = string.Format( "{0}", 1 ) + string.Format( "{0}", 1 );
```

giggio:
Closes #690

### DIFF
--- a/src/CSharp/CodeCracker/CodeCracker.csproj
+++ b/src/CSharp/CodeCracker/CodeCracker.csproj
@@ -81,6 +81,7 @@
     <Compile Include="Reliability\UseConfigureAwaitFalseCodeFixProvider.cs" />
     <Compile Include="Style\ConsoleWriteLineAnalyzer.cs" />
     <Compile Include="Style\ConsoleWriteLineCodeFixProvider.cs" />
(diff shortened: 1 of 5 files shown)


carloscds:
@giggio Fix is correcting only one String.Format() each time. Is it correct ?

giggio:
@carloscds The fix all should be fixing all of them. The simple fix will fix only one, or the other.

carloscds:
@giggio Not fix all, only fix. When I click to fix, only first segment is fixed. Is it expect result ?

giggio:
@carloscds yes.

Issue #698, opened 2016-02-01
BUG: CC0006 (foreach) edge case for non-writable struct field


jakubsuchybio:
Hi.
So I found another one. But this one is interesting.
Here is a repro code: (tested with 1.0.0-rc5-1063)
```c#
		struct Foo
		{
			public int X;
		}

		static void Goo()
		{
			var array = new Foo[1];
			for( int i = 0; i < array.Length; i++ )
			{
				var actual = array[i];
				array[i].X = 5;
			}
		}
```
Here this flags for with CC0006 to convert it to foreach, but when it does then its uncompileable, because of writing into struct temp var doesn't work.

But when you delete the line with
```c#
var actual = array[i];
```
Then it works ok and doesn't flag it.

So I think that it doesn't matter to look for local var assignment of indexed array, when for is dealing with struct foreach temp variable.

Also, fix for lists and other enumerables. For example:

````csharp
var list = new List<Foo>();
for (int i = 0; i < list.Count; i++)
{
    var actual = list[i];
    actual.X = 5;
}
````

giggio:
Bug confirmed, it should not raise a diagnostic.

PR #705, opened 2016-02-02
Fix Issue #698

carloscds:


src/CSharp/CodeCracker/Style/ForInArrayAnalyzer.cs
```diff
@@ -65,7 +65,18 @@ private static void Analyzer(SyntaxNodeAnalysisContext context)
                     return !identifierSymbol.Equals(arrayId);
                 });
             if (otherUsesOfIndexToken != 0) return;
-
+            var baseTypeName = (SymbolType(arrayId) as IArrayTypeSymbol)?.ElementType?.BaseType?.Name ?? "";
```
giggio: You are assuming the enumerable is an array. It could be any type that supports the enumerable pattern. It does not even need to be an `IEnumerable`.

carloscds: Yes, but in this case, all types for this analyzer are Enumerable (IsEnumerable function), even so, are an array.

giggio:
@carloscds This still triggers the diagnostic:

````csharp
var list = new List<Foo>();
for (int i = 0; i < list.Count; i++)
{
    var actual = list[i];
    actual.X = 5;
}
````

You can't rely on the enumerable type, it does not have to be an array, or even an IEnumerable. You have to look into the symbol of the `actual` variable, and get its type.

giggio:
This still fails, it is still raising a diagnostic:

````csharp
class Foo
{
    void Bar()
    {
        var list = new MyList();
        for (int i = 0; i < list.Count; i++)
        {
            var actual = list[i];
            actual.X = 5;
        }
    }
}
struct Struct
{
    public int X { get; set; }
}
class MyList
{
    public int Count => 1;
    public Struct this[int index] => new Struct();
    public System.Collections.Generic.IEnumerator<Struct> GetEnumerator()
    {
        yield return new Struct();
    }
}
````

The solution is to not look into the array type symbol, but to the symbol for the iteration variable. In the example above, look for the symbol for the `actual` variable, and if it is a value type, we don't raise the diagnostic.

carloscds:
@giggio Ok, I agree, but if I change based on this, this analyzer will not work for ValueType arrays, because all variables initialized into for() also will be ValueTypes, even if it will not assigned.

carloscds:
@giggio I changed the code for you look at fails tests.

carloscds:
@giggio Done!

PR #758, opened 2016-03-31
Can copy event to variable without crashing when inside a statement

### LINKED ISSUE #735: BUG: CC0016 "CopyEventToVariableBeforeFireCodeFixProvider" crashes
This code resulted in `CopyEventToVariableBeforeFireCodeFixProvider` crash.
```Csharp
	public Func<AToken, bool> AllowInteraction { get; protected set; }

	protected bool AllowedInteraction()
	{
		if (!AllowInteraction(this) || InteractionType == TokenInteractionType.Preassembled)
		{
			// ...
			return false;
		}
		return true;
	}
```

Expected fix:

```Csharp
	public Func<AToken, bool> AllowInteraction { get; protected set; }

	protected bool AllowedInteraction()
	{
		var e = AllowInteraction;
		if (e != null)
		{
			if (!e(this) || InteractionType == TokenInteractionType.Preassembled)
			{
				// ...
				return false;
			}
		}
		return true;
	}
```

giggio:
Fixes #735

### DIFF
--- a/src/CSharp/CodeCracker/Design/CopyEventToVariableBeforeFireCodeFixProvider.cs
+++ b/src/CSharp/CodeCracker/Design/CopyEventToVariableBeforeFireCodeFixProvider.cs
@@ -30,12 +30,13 @@ namespace CodeCracker.CSharp.Design
             return Task.FromResult(0);
         }
 
-        private async static Task<Document> CreateVariableAsync(Document document, Diagnostic diagnostic, CancellationToken ct)
+        private async static Task<Document> CreateVariableAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
         {
-            var root = await document.GetSyntaxRootAsync(ct).ConfigureAwait(false);
+            var root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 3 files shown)


carloscds:
@giggio When I have a code like this:

'''''csharp
public class foo
    {
        public Func<AToken, bool> AllowInteraction { get; protected set; }
        public Func<AToken, bool> SecondInteraction { get; protected set; }

        protected bool AllowedInteraction()
        {
            if (!AllowInteraction(this) || InteractionType == TokenInteractionType.Preassembled || !SecondInteraction(this))
            {
                // ...
                return false;
            }
            return true;
        }
    }
'''''

I have an incorrect fix after second fix:
'''''csharp
public class foo
    {
        public Func<AToken, bool> AllowInteraction { get; protected set; }
        public Func<AToken, bool> SecondInteraction { get; protected set; }

        protected bool AllowedInteraction()
        {
            var handler = AllowInteraction;
            if (handler != null)
            {
                var handler = SecondInteraction;
                if (handler != null)
                    if (!handler(this) || InteractionType == TokenInteractionType.Preassembled || !handler(this))
                    {
                        // ...
                        return false;
                    }
            }

            return true;
        }
    }

'''''

giggio:
@carloscds updated.

PR #800, opened 2016-06-06
Suggest default in switch statement  

### LINKED ISSUE #780: CC0120: Suggest default for switch statements (C#)
A lot of bugs are caused by non-exhaustive matching in switch statements.
If there is no default case one should be added that throws an exception.

Before:

``` csharp
switch (a)
{
    case "":
        break;
}
```

After:

``` csharp
switch (a)
{
    case "":
        break;
    default:
        throw new Exception("Unexpected Case");
}
```

Should not raise for `boolean` if `true` and `false` are already handled:

``` csharp
var f = true;
switch (f)
{
    case true:
        break;
    case false:
        break;
    default:
        break;
}
```

Should not raise a diagnostic if [CS0151](https://msdn.microsoft.com/en-us/library/dh0x1stb.aspx)(C# A switch expression or case label must be a bool, char, string, integral, enum, or corresponding nullable type) is already raised in the switch.

Diagnostic Id: `CC0120`
Category: `Design`
Severity: `Warning`

Same as #767, but for C#.

@cezarant is working on it.

cezarant:
Fixes #780
CC0120: Suggest default for switch statements (C#)

This is my first pull request, if there is a problem , like a feedback. 
Thank you!

### DIFF
--- a/src/CSharp/CodeCracker/Design/SwitchWithoutDefaultAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Design/SwitchWithoutDefaultAnalyzer.cs
@@ -1,0 +1,61 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+
+namespace CodeCracker.CSharp.Design
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class SwitchWithoutDefaultAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string Title = "Your Switch maybe include default clause";
+        internal const string MessageFormat = "{0}";
+        internal const string Category = SupportedCategories.Design;
+        internal static readonly DiagnosticDescriptor Rule = new DiagnosticDescriptor(
+            DiagnosticId.SwitchCaseWithoutDefault.ToDiagnosticId(),
+            Title,
+            MessageFormat,
+            Category,
+            DiagnosticSeverity.Warning,
+            isEnabledByDefault: true,
+            helpLinkUri: HelpLink.ForDiagnostic(DiagnosticId.SwitchCaseWithoutDefault));
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context) => context.RegisterSyntaxNodeAction(Analyzer, SyntaxKind.SwitchStatement);
+
+        private static void Analyzer(SyntaxNodeAnalysisContext context)
+        {
+            if (context.IsGenerated()) return;
+            // checks if the compile error CS0151 it has fired.If Yes, don't report the diagnostic
+            var diagnostics = from dia in context.SemanticModel.GetDiagnostics()
+                              where dia.Id == "CS0151"
+                              select dia;
+            if (diagnostics.ToList().Count == 0)
+            {
+                var defaultSwitchSyntaxe = (SwitchStatementSyntax)context.Node;
+                if (defaultSwitchSyntaxe.DescendantNodes().Where(n => n.Kind() == SyntaxKind.DefaultSwitchLabel).ToList().Count == 0)
+                {
+                    var variables = ((LocalDeclarationStatementSyntax)defaultSwitchSyntaxe.GetPreviousStatement()).Declaration.Variables.Last();
+                    if (variables != null)
+                    {
+                        // if boolean variables, don't report the diagnostic
+                        var kindVariable = ((VariableDeclaratorSyntax)variables).Initializer.Value.Kind();
+                        if ((kindVariable != SyntaxKind.FalseLiteralExpression) && (kindVariable != SyntaxKind.TrueLiteralExpression))
+                        {
+                            var diagnostic = Diagnostic.Create(Rule, defaultSwitchSyntaxe.GetLocation(), "Consider put an default clause in Switch.");
+                            context.ReportDiagnostic(diagnostic);
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


src/CSharp/CodeCracker/Design/SwitchWithoutDefaultAnalyzer.cs
```diff
@@ -0,0 +1,61 @@
...
+        public override void Initialize(AnalysisContext context) => context.RegisterSyntaxNodeAction(Analyzer, SyntaxKind.SwitchStatement);
+
+        private static void Analyzer(SyntaxNodeAnalysisContext context)
+        {
+            if (context.IsGenerated()) return;
+            // checks if the compile error CS0151 it has fired.If Yes, don't report the diagnostic
+            var diagnostics = from dia in context.SemanticModel.GetDiagnostics()
+                              where dia.Id == "CS0151"
+                              select dia;
+            if (diagnostics.ToList().Count == 0)
```
thorgeirk11: Would simplify this down to:
```CSharp 
if (!diagnostics.Any())

File: test/CSharp/CodeCracker.Test/Design/SwicthWithoutDefaultTests.cs [hunk 2]
```diff
@@ -0,0 +1,128 @@
...
+                                    ConsoleApplication1 
+                                    {
+                                        class TypeName
+                                        {
+                                              static void Main()
+                                              {
+                                                 var s = false;
+                                                 switch (s)
+                                                 {
+                                                  case false;break:{break;}
```
thorgeirk11: This has compile issues:

Did you mean:
```CSharp
var s = false;
switch (s)
{
	case false:
		break;
}

File: src/CSharp/CodeCracker/Design/SwitchWithoutDefaultAnalyzer.cs [hunk 3]
```diff
@@ -0,0 +1,61 @@
...
+            {
+                var defaultSwitchSyntaxe = (SwitchStatementSyntax)context.Node;
+                if (defaultSwitchSyntaxe.DescendantNodes().Where(n => n.Kind() == SyntaxKind.DefaultSwitchLabel).ToList().Count == 0)
+                {
+                    var variables = ((LocalDeclarationStatementSyntax)defaultSwitchSyntaxe.GetPreviousStatement()).Declaration.Variables.Last();
+                    if (variables != null)
+                    {
+                        // if boolean variables, don't report the diagnostic
+                        var kindVariable = ((VariableDeclaratorSyntax)variables).Initializer.Value.Kind();
+                        if ((kindVariable != SyntaxKind.FalseLiteralExpression) && (kindVariable != SyntaxKind.TrueLiteralExpression))
```
thorgeirk11: This method does not cover this case:

```CSharp
using System; 
class Foo
{
	void Bar(bool myBool)
	{
		switch (myBool)
		{
			case false:
				break;
		}
	}
}

**Comment by cezarant** [comment]
Thanks for the feedback @thorgeirk11 ! 
I will make the changes indicated :)

**Comment by carloscds** [comment]
Hi @cezarant, this code not raise analyzer:

```csharp
var s = false;
            switch (s)
            {
                case false:
                    {
                        break;
                    }
            }
```

giggio:
There are several small problems, @cezarant, are you going to move forward with this PR?

giggio:
@carloscds 

Actually, this code:
````csharp
var s = false;
            switch (s)
            {
                case false:
                    {
                        break;
                    }
            }
````

Should raise a diagnostic, as it does not have a `case true`. We can't go on trying to verify if `s` had an assignment, this would make this analyzer impossible to finish. We should keep the analysis to the `switch` statement, IMO.

test/CSharp/CodeCracker.Test/Design/SwicthWithoutDefaultTests.cs
giggio: Right, it has to be fixed.

cezarant:
@giggio I will study more about the Code Cracker. I did not think I'd find so many difficulties in carrying out this analyzer. Forgive me for doing pull Request with a low coverage tests. I will be more cautious next time.

carloscds:
@cezarant  Thanks for your collaboration and go ahead!

carloscds:
@cezarant Do you have some news about this ?

cezarant:
Hi @carloscds I was waiting for feedback for Fix #808, but I can work here. Following the above suggestions. I think I have to make a rebasing of the code that was loaded. Right?

carloscds:
@cezarant We can process this PR separately.

cezarant:
Hi devs! Anybody here have had problems to build the project, when change of machine? 
I receive [this error](https://github.com/dotnet/roslyn/issues/8077) (VSSDK1031) when i build.  
I am making several attempts to resolve the error, but if someone has been there and can help me, I would appreciate

carloscds:
@cezarant Here project is compiling. Do you check for any extension installed on last days ?

cezarant:
Hello @carloscds  I did a format in my notebook, and re-install the Visual Studio. It is the only abnormal thing I did.  Now, i unmarked options on the VSIX tab, and seems to have settled. 

vsix
If no IDE error happen more, I will upload the changes soon.

carloscds:
@cezarant CodeCracker users VSIX project.

cezarant:
@carloscds  I had two problems with the original branch that gave rise to this PR:
- The build generate the error VSSDK1031, and only works when the retreat option to "Create VSIX during the build"
- The rebase is taking too long to be done 

So, I clone the actual version of the code cracker, created a new branch and put the changes.
The new branch is able to merge, I wonder if you see any problem if I generate another pull request for this issue.  
I did not want to do that. But it is giving a lot of work to restore the old branch.

carloscds:
@cezarant This branch has some conflicts, can you update it ?

cezarant:
@carloscds Being my first analyzer, the original branch of this pull request has unnecessary modifications to files that were not part of the issue. I created a new pull request, starting from the beginning and already contemplating the suggestions mentioned above.

It is a bad idea?

carloscds:
@cezarant Look: 
image

cezarant:
Hello @carloscds i resolved the conflicts in the the branch that I used to create this pull request. 
However, as this branch is quite old (06  Jun) , the comparison with the current code generated 217 modified files. Even though I made the rebasing. 

I feel unsafe to perform the pull request that branch. I can make a new rebasing, 
to reduce the number of modified files, but I think it would be better to close this pull request, and to continue on the PR #851. 
Because in addition to having had an increase in test coverage, PR #851 had a much smaller number of modified files: only six. 

I'm sorry to have created another pull request, but could not find another way to check test coverage.

carloscds:
@cezarant If you will create a new PR, please close this.

PR #805, opened 2016-07-04
CC0122 Replace Task.Result with await Task

### LINKED ISSUE #734: Replace Task.Result with await Task
**New analyzer**: Replace `Task.Result` with `await` and make method `async Task`

Recently we've found some deadlocks in our code. They were caused by a stray `.Result` in situations when there was an `async` method earlier in the call stack. [Using `.Result` is considered a bad practice](https://msdn.microsoft.com/en-us/magazine/jj991977.aspx). 
I propose an analyzer that removes `.Result` and replaces it with an `await ` call

Before:

````csharp
private bool doSomething()
{
    return asyncMethod().Result;
}
````

After:

````csharp
private Task<bool> doSomething()
{
    return await asyncMethod();
}
````

Category: Reliability or Design
Severity: By default hidden. I would probably change it to Warning in my own code

I'm curious what's your opinion on
1. This refactoring will cause a build error, because method's signature changes
2. Are there any use cases where using `.Result` is OK?

AmadeusW:
Fixes #734 
**CC0122 Replace Task.Result with await Task**

Adds Design Warning CC0122 when `System.Threading.Tasks.Task<TResult>.Result` is used in `async` method.
The code fix provider replaces `.Result` with `async `

Screenshot of a non trivial case:
cc0122


giggio:
@AmadeusW You have some problem with your commits, you probably merged from CodeCracker's master into your branch, or something else. You can see that on the `Commit` tab here at github, it shows `12` commits and `35` changed files.
Also, you should change the pull to branch `master`, not `vnext`.

giggio:
Taking a look at the commits, what you should do is:

* Update your master branch, according to CodeCracker's master branch `git checkout master`, then `git pull https://github.com/code-cracker/code-cracker.git master`.
* Create a new branch `git checkout -b newbranch`
* Cherry pick your commits into it `git cherry-pick 91e3e6d4fa05ea002c1844ab2665a00d0f8ca18c` and `git cherry-pick 23ebf1b89136cee55fd48caa48c0949dce0cd848`.

Then open the PR with this branch, and close this one.

AmadeusW:
Thanks @giggio, I'm closing this in favor of #820

PR #806, opened 2016-07-08
Fix 'GC.SuppressFinalize within the finally block' #776

### LINKED ISSUE #776: BUG: GC.SuppressFinalize within any block (CC0029)
`public override void Dispose()
{
	try
	{
		if (MyConnection != null)
			(MyConnection.Dispose();
	}
	finally
	{
		GC.SuppressFinalize(this);
	}
}
`

When you add a `GC.SuppressFinalize (this)` in a finally block the codecrack suggests the addition of `SuppressFinalize` method. I believe this is a small bug, or not?

code

Should work with any block too:

This should not raise a diagnostic:

````csharp
public void Dispose()
{
    if (condition)
        GC.SuppressFinalize(this);
}
````

vs-krasnov:
Fixes #776 

DisposablesShouldCallSuppressFinalizeAnalyzer did not find GC.SuppressFinalize(this) in finally block. 
I added proper test DoNotWarnIfClassImplementsIDisposableWithSuppressFinalizeCallInFinally() to check this issue.

### DIFF
--- a/src/CSharp/CodeCracker/Usage/DisposablesShouldCallSuppressFinalizeAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Usage/DisposablesShouldCallSuppressFinalizeAnalyzer.cs
@@ -1,4 +1,4 @@
-﻿using System;
+﻿using System.Collections.Generic;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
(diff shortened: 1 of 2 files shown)


vs-krasnov:
@giggio Could you please help me with AppVeyor? I've got a strange error - "Error uploading artifact to Google Storage." Does it connect to AppVeyor's settings or related to my commit?

giggio:
@vs-krasnov I have asked for a rebuild on AppVeyor, the build worked fine.

giggio:
There is a problem, if there is no `finally`, there is a null reference exception.
But I think we should do better. First, we should check for a finally block. i don't think we will be able to make sure that the user is doing the correct thing. What if the SuppressFinalize is inside an `if`? The programmer should know better, not us, IMO. This, right now, also generates a warning:
````csharp
public void Dispose()
{
    if (condition)
        GC.SuppressFinalize(this);
}
````
I suggest we simply search for this statement anywhere. So, statements should be:

````csharp
var statements = method.Body?.DescendantNodes.OfType<ExpressionStatementSyntax>();
````

Would you update the fix and add some tests for this new conditions? I have already updated the description on #776.

BTW, nice refactoring.

Also, please PR into branch `v1.0.x`, not `master`.

vs-krasnov:
@giggio Thanks for your help with AppVeyor and advice on improving the fix. Sure, I will update the commit, sorry for PR into `master`.

vs-krasnov:
Created new PR #819 to `v1.0.x`

PR #815, opened 2016-08-09
Enable CC0029 analyzer to work with arrow methods

### LINKED ISSUE #809: BUG: GC.SuppressFinalize with arrow methods (CC0029)
This should not raise `CC0029`:

````csharp
    class Foo : IDisposable
    {
        public void Dispose() => GC.SuppressFinalize(true);
    }
````

Right now it does not know how to work with arrow methods. Should update the analyzer and the fix.

giggio:
Fixes #809


carloscds:
@giggio This PR has conflitcs, can you check ?

giggio:
@carloscds updated!

PR #828, opened 2016-08-22
Fix #822. Ignore when IsMath() called from field member.

### LINKED ISSUE #822: Bug: "UseStaticRegexIsMatchAnalyzer" causes an exception (CC0081)
Hi,

while working with regular expressions in C#, an exception occured in the error list caused by code-cracker.

After some playing around with the code, i found the suspicous part in my code which is causing the exception.

````csharp
public class RegexTestClass
{
    private TestModel testModel;

    private void Test(string text)
    {
        if (testModel.Regex.IsMatch(text))
        {
            return;
        }
    }
}

public class TestModel
{
    public System.Text.RegularExpressions.Regex Regex { get; set; }
}
````

The exception (only got it in german):

````
Analyzer 'CodeCracker.CSharp.Performance.UseStaticRegexIsMatchAnalyzer' threw 
an exception of type 'System.InvalidCastException' with message
'Unable to cast object of type 'Microsoft.CodeAnalysis.CSharp.Syntax.MemberAccessExpressionSyntax' to type 'Microsoft.CodeAnalysis.CSharp.Syntax.IdentifierNameSyntax'.'.	
````

A more complex scenario:

````csharp
public class RegexTestClass
{
    private C c;

    private void Test(string text)
    {
        if (c.TestModel.Regex.IsMatch(text))
        {
            return;
        }
    }
}
public class C
{
    public TestModel TestModel { get; set; }
}
public class TestModel
{
    public System.Text.RegularExpressions.Regex Regex { get; set; }
}
````

````
Active	Analyzer 'CodeCracker.CSharp.Performance.UseStaticRegexIsMatchAnalyzer' threw the following exception:
'Exception occurred with following context:
Compilation: ConsoleApplication1
SyntaxTree: c:\temp\ConsoleApplication1\ConsoleApplication1\Program.cs
SyntaxNode: testModel.Regex.IsMatch(text) [InvocationExpressionSyntax]@[439..468) (21,16)-(21,45)

System.InvalidCastException: Unable to cast object of type 'Microsoft.CodeAnalysis.CSharp.Syntax.MemberAccessExpressionSyntax' to type 'Microsoft.CodeAnalysis.CSharp.Syntax.IdentifierNameSyntax'.
   at CodeCracker.CSharp.Performance.UseStaticRegexIsMatchAnalyzer.Analyzer(SyntaxNodeAnalysisContext context) in C:\p\code-cracker\src\CSharp\CodeCracker\Performance\UseStaticRegexIsMatchAnalyzer.cs:line 44
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.<>c__DisplayClass42_1`1.<ExecuteSyntaxNodeAction>b__1()
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.ExecuteAndCatchIfThrows_NoLock(DiagnosticAnalyzer analyzer, Action analyze, Nullable`1 info)
````

I guess the problematic line is in [here](https://github.com/code-cracker/code-cracker/blob/master/src/CSharp/CodeCracker/Performance/UseStaticRegexIsMatchAnalyzer.cs) on line 44

Thanks, and thanks again for fixing the other issue I submitted!

The PR has to go on the `v1.0.x` branch.

carloscds:


giggio:
Carlos, the branch is incorrect, it should go to `v1.0.x`.
Also you are optimizing for the simplest case. You have to eliminate all possibility of this cast fail, or search for the identifier. You could loop until the `Expression` is a `MemberAccessExpression`.

But we can be better than that.

Check this 2 lines that follow:

````csharp
var variableSymbol = context.SemanticModel.GetSymbolInfo(((IdentifierNameSyntax)syntaxNode).Identifier.Parent).Symbol;
if (variableSymbol?.Kind != SymbolKind.Local) return
````

We exit if the variable is not local. If it is a member access, it will never be local.
So, checking for:

````csharp
if (!(memberExpression.Expression is IdentifierNameSyntax)) return;
````

Should solve it.

For example, this case still throws (I added another class.):

````csharp
public class RegexTestClass
{
    private C c;

    private void Test(string text)
    {
        if (c.TestModel.Regex.IsMatch(text))
        {
            return;
        }
    }
}
public class C
{
    public TestModel TestModel { get; set; }
}
public class TestModel
{
    public System.Text.RegularExpressions.Regex Regex { get; set; }
}
````

carloscds:
Close to point to branch v1.0.x

giggio:
Superseeded by #832.

PR #836, opened 2016-09-05
Fix774

### LINKED ISSUE #774: BUG: CC0030 should not try to make a pointer const
Fix does not work for 'const' pointers.
cc0030

carloscds:
Fix #774


giggio:
@carloscds there was a problem with the PR, I cherry-picked your commit and it is now on `v1.0.x`. I am closing this PR as the code is now there.

PR #849, opened 2016-10-19
#848 - "Disposable Field Not Disposed" rule does not recognize null p…

### LINKED ISSUE #848: "Disposable Field Not Disposed" rule does not recognize null propagation
# Bug

"Disposable Field Not Disposed" rule does not recognize null propagation

**This works:**

``` csharp
if (disposableField != null)
                disposableField.Dispose();
```

**This doesn't:**

``` csharp
disposableField?.Dispose();
```

 **Unit test:**

``` csharp
[Fact]
        public async Task WhenUsingTheDisposablePatternWithNullPropagationItDoesNotCreateDiagnostic()
        {
            const string source = @"
using System;
using System.IO;
public class A : IDisposable
{
    private MemoryStream disposableField = new MemoryStream();
    public void Dispose()
    {
        Dispose(true);
        GC.SuppressFinalize(this);
    }
    private void Dispose(bool disposing)
    {
        if (disposing)
            disposableField?.Dispose();
    }
}";
            await VerifyCSharpHasNoDiagnosticsAsync(source);
        }
```

Nozziel:
Fixes #848 and includes additional unit test

### DIFF
--- a/src/CSharp/CodeCracker/Usage/DisposableFieldNotDisposedAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Usage/DisposableFieldNotDisposedAnalyzer.cs
@@ -92,13 +92,41 @@ namespace CodeCracker.CSharp.Usage
             var hasDisposeCall = body.DescendantNodes().OfKind<InvocationExpressionSyntax>(SyntaxKind.InvocationExpression)
                 .Any(invocation =>
                 {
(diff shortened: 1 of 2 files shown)


giggio:
This was supposed to have gone to branch `v1.0.x`. I just cherry picked it with commit cfa4fa358c3b0622fb2f06741461f041802bc55f.

PR #851, opened 2016-10-28
- Fix #780 - Adding more tests and and solve conflicts 

### LINKED ISSUE #780: CC0120: Suggest default for switch statements (C#)

cezarant:
Following the suggestions of the developers in the first PR to resolve the issue # 780, this update has tests for situations of Boolean and integer parameters and treatment of exception CS0151.

### DIFF
--- a/src/CSharp/CodeCracker/CodeCracker.csproj
+++ b/src/CSharp/CodeCracker/CodeCracker.csproj
@@ -46,6 +46,8 @@
     <Compile Include="Design\InconsistentAccessibility\InconsistentAccessibilityInPropertyType.cs" />
     <Compile Include="Design\MakeMethodStaticAnalyzer.cs" />
     <Compile Include="Design\MakeMethodStaticCodeFixProvider.cs" />
+    <Compile Include="Design\SwitchWithoutDefaultAnalyzer.cs" />
+    <Compile Include="Design\SwitchWithoutDefaultCodeFixProvider.cs" />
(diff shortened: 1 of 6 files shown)


giggio:
Found bug #859 on this code. Maybe @cezarant wants to fix it?

cezarant:
Yes, @giggio  I'm already checking.

giggio:
@cezarant I fixed, also there was many refactorings. Check out e1c18b6

cezarant:
Ok @giggio. I will study the refactorings. To learn more about it

PR #861, opened 2017-01-08
Fixes #798

### LINKED ISSUE #798: BUG: CC0014 converts if y then x += 1 else x =1 to x +=(if(y, 1, 1)
# Bug

In cases where one branch of an if increments a variable and the other assigns a value to it then CC0014 should not offer a fix.

``` vb.net
Dim s As String
If True Then
    s += "a"
Else
    s = "b"
End If
```

Current output after fix applied:

``` vb.net
s += If(True, "a", "b")
```

Fix should go to branch `v1.0.x`.

@bergarces is working on it.

bergarces:
Fixes #798

The original code was able to handle a combination of simple and non-simple assignment operators if the simple one was on the if and the complex one on the else. I have amended the code so that the opposite case can be handled as well. The changes should also work if there are two different non-simple assignment operators on if and else.

Tests have been added for all the described scenarios.

I've encountered an issue whilst applying the fix for which I don't have a clear answer. On line 140 I wasn't able to call SyntaxFactory.Token(SyntaxKind.SimpleAssignmentStatement) as it would throw a System.ArgumentOutOfRangeException, but it builds and passes the tests with an EqualsToken. I'm not sure why this is the case.

Please let me know if any further amendment is needed.

Thanks.

### DIFF
--- a/src/VisualBasic/CodeCracker/Style/TernaryOperatorCodeFixProviders.vb
+++ b/src/VisualBasic/CodeCracker/Style/TernaryOperatorCodeFixProviders.vb
@@ -112,7 +112,7 @@ Namespace Style
                 EnsureNothingAsType(semanticModel, type, typeSyntax).
                 ConvertToBaseType(elseType, type)
 
-            If ifAssign.OperatorToken.Text <> "=" Then
+            If ifAssign.OperatorToken.Text <> "=" AndAlso ifAssign.OperatorToken.Text = elseAssign.OperatorToken.Text Then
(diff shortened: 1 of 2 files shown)


giggio:
Good way to solve the problem! We were actually asking for disabling the analyzer in those scenarios, you got them to work. Nice job!
I just removed the unnecessary `allowNewCompilerDiagnostics` and added the tests with `+`, as #798 stated, and they were missing on commit 5020872.

PR #874, opened 2017-02-06
Fix #872: UnusedParametersAnalyzer

### LINKED ISSUE #872: CC0057 UnusedParametersAnalyzer should not be triggered on virtual methods
## Description

CC0057 gets triggered on virtual methods. In my opinion virtual methods don't need to use all the parameters available in the base class implementation, because the parameters serve just as interface for future extensibility in subclasses.

## Example: 

In my case the virtual method is just a dummy implementation that doesn't perform anything at all. Subclasses however might override the method and actually use the parameter.

````csharp
internal abstract class BaseClass
{

  // ...

  protected virtual void PreProcess(MyDataStructure ds)
  {
     // base implementation: don't do anything. Subclasses might override the method and implement custom behavior
  }
}
````

### Current output:

Warning CC0057 "UnusedParametersAnalyzer": Parameter "ds" not used.

### Expected output:

Occurrence should be ignored, because method is virtual.

StefanTheiner:
CC0057 UnusedParametersAnalyzer should not be triggered on virtual methods
Includes test case and fix.

Fixes #872

### DIFF
--- a/src/CSharp/CodeCracker/Usage/UnusedParametersAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Usage/UnusedParametersAnalyzer.cs
@@ -59,6 +59,12 @@ namespace CodeCracker.CSharp.Usage
             }
 
             var method = methodOrConstructor as MethodDeclarationSyntax;
+
+            // It is legit for virtual methods to have parameters that aren't used in the default
+            // base class implementation, but are only provided for sub classes instead.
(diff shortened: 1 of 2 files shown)


dnickless:
I strongly support this change and would love to see it merged. Mind you, it could be enhanced in a second version to scan overridden methods (if no other class overrides the offending virtual method or no overridden method makes use of the offending parameter then we would still want to flag this case).

carloscds:
@dnickless Merged!

PR #904, opened 2017-03-22
New Analyzer CC0125 - getter only auto property

### LINKED ISSUE #881: Replace getter only properties with backing readonly field with getter-only auto-property
#New analyzer:

I published an analyzer [here](https://github.com/ifpanalytics/Ifp.Analyzers) that 
finds getter only properties with backing readonly field and converts them to getter-only auto-properties.
Is there interest to port this one over to this library?

Before:

```cs
    class AwesomeService
    {
        private readonly ILogger _Logger;
        public AwesomeService(ILogger logger)
        {
            _Logger = logger;
        }

        protected ILogger Logger
        {
            get
            {
                return _Logger;
            }
        }
    }
```

After:

```cs
    class AwesomeService
    {
        public AwesomeService(ILogger logger)
        {
            Logger = logger;
        }

        protected ILogger Logger { get; }
    }
```

See [here](https://github.com/ifpanalytics/Ifp.Analyzers) for further details.

Diagnostic Id: `CC0125`
Category: `Refactoring`
Severity: `Hidden`

MaStr11:
Fixes #881.

New analyzer CC0125 - getter only auto property.

Port from [https://github.com/ifpanalytics/Ifp.Analyzers](https://github.com/ifpanalytics/Ifp.Analyzers).

Remaining issues:
* **Renaming may introduce name clash**. See Test method [RenamingOfFieldAccessCanIntroduceNameClashesCaughtByCompilerWarningCS1717](https://github.com/MaStr11/code-cracker/blob/CC0125GetterOnlyAutoProperty/test/CSharp/CodeCracker.Test/Refactoring/ReplaceWithGetterOnlyAutoPropertyTests.cs#L480)
* **Batchfixer** The CodeFixProvider uses the `WellKnownFixAllProviders.BatchFixer`. The `BatchFixer` is not able to fix multiple properties in a class at once (fixes about every second property in a class at once). There are not tests covering the `FixAllProvider`.

Let me know if any of these issues are show-stoppers and I will look into it.

I also prepared the documentation for the analyzer at [https://github.com/code-cracker/code-cracker.github.io/compare/master...MaStr11:DocsForCC0125](https://github.com/code-cracker/code-cracker.github.io/compare/master...MaStr11:DocsForCC0125).
I can create a pull request as needed.


MaStr11:
@carloscds The last commit (fa20c95) only added one line of code to prevent an ArgumentException and to increase performance.

MaStr11:
@carloscds I pushed another small change related to the previous bug. The cause of the bug is now covert by unit tests. The fixer does not rename fields in partial classes causing a [CS0103 ](https://msdn.microsoft.com/en-US/library/t8zbaa6f.aspx) `The name 'identifier' does not exist in the current context` compiler error after the fix is applied.

Issue #906, opened 2017-03-28
TernaryOperatorWithReturnCodeFixProvider NullReferenceException


theraot:
I am getting a NullReferenceException in Visual Studio, the following is the stack pile (Spanish):

```
System.NullReferenceException : Referencia a objeto no establecida como instancia de un objeto.
   at CodeCracker.CSharp.Style.TernaryOperatorCodeFixHelper.CastToBaseType(ExpressionSyntax ifExpression,ITypeSymbol ifType,ITypeSymbol elseType,ExpressionSyntax trueExpression)
   at CodeCracker.CSharp.Style.TernaryOperatorCodeFixHelper.CreateExpressions(ExpressionSyntax ifExpression,ExpressionSyntax elseExpression,ITypeSymbol ifType,ITypeSymbol elseType,ITypeSymbol ifConvertedType,ITypeSymbol elseConvertedType,TypeSyntax typeSyntax,SemanticModel semanticModel,ExpressionSyntax& trueExpression,ExpressionSyntax& falseExpression)
   at CodeCracker.CSharp.Style.TernaryOperatorCodeFixHelper.CreateExpressions(ExpressionSyntax ifExpression,ExpressionSyntax elseExpression,SemanticModel semanticModel,ExpressionSyntax& trueExpression,ExpressionSyntax& falseExpression)
   at async CodeCracker.CSharp.Style.TernaryOperatorWithReturnCodeFixProvider.MakeTernaryAsync(<Parámetros desconocidos>)
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.GetChangedSolutionAsync(<Parámetros desconocidos>)
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.ComputeOperationsAsync(<Parámetros desconocidos>)
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.ComputePreviewOperationsAsync(<Parámetros desconocidos>)
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.GetPreviewOperationsAsync(<Parámetros desconocidos>)
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedAction.GetPreviewResultAsync(<Parámetros desconocidos>)
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedActionWithNestedFlavors.<>c__DisplayClass11_0.<GetPreviewAsync>b__0(<Parámetros desconocidos>)
[...]

PR #909, opened 2017-03-29
Fix906

carloscds:
Fix bug on ternary operator. #906


giggio:
@carloscds The test passes even if there are no changes made to `src/CSharp/CodeCracker/Style/TernaryOperatorCodeFixProvider.cs`. Would you please verify?

PR #938, opened 2017-06-21
CC0014 casting rules bugs

### LINKED ISSUE #911: BUG: CC0014 Casting to interface or implicit casts for the ternary operator are fixed wrong
The code fix for CC0014 tries to detect a common base type for both sides of the ternary operator.
This fails under certain circumstances.

## Target of the assignment is an interface:

````csharp
IComparable c = null;
if (cond)
    c = 1;
else
    c = "Text";
````

Current output after fix applied:

````csharp
IComparable c = null;
c = cond ? (object)1 : "Text";
````

Expected output after fix applied:

````csharp
IComparable c = null;
c = cond ? (IComparable)1 : "Text";
```` 

## Implicit conversions:

````csharp
double d;
if (cond)
    d = 1;
else
    d = 1.1;
````

Current output after fix applied:

````csharp
double d;
d = cond ? (ValueType)1 : 1.1;
````

Expected output after fix applied:

````csharp
double d;
d = cond ? 1 : 1.1;
```` 

## Bug fix

To fix this bug I would propose to change the logic, that finds the common base type of the `true` and `false` hand side of the ternary operator. I would suggest to take the target type into consideration.

For the `TernaryOperatorWithAssignmentCodeFixProvider` the target type is the type of the variable. This type is accessible at this [line of code](https://github.com/code-cracker/code-cracker/blob/master/src/CSharp/CodeCracker/Style/TernaryOperatorCodeFixProvider.cs#L76) via `semanticModel.GetTypeInfo(assignmentExpressionInsideIf.Left).ConvertedType`

For the `TernaryOperatorWithReturnCodeFixProvider` the target type is the return type of the method/property. This type is a accessible [here](https://github.com/code-cracker/code-cracker/blob/master/src/CSharp/CodeCracker/Style/TernaryOperatorCodeFixProvider.cs#L41) via `semanticModel.GetTypeInfo(returnStatementInsideIf.Expression).ConvertedType`

The CodeFixer could be changed to apply the conversion based on these type informations. If the bug is confirmed and you agree with the proposal I could start working on the bug fix.

MaStr11:
Fixes #911.


giggio:
@MaStr11 Good job, a simple solution. Well done.
The only comment I have is that you requested a pull to the wrong branch, it should had been to `v1.0.x`. The branch `master` is for new analyzers and improvements, fixes always go to the current fix branch, which is `v1.0.x` now. I will cherry-pick it back there so the next version receives the update.

PR #939, opened 2017-06-23
CC0118 Remove to string only if it is save to do

### LINKED ISSUE #866: BUG: CC0118 - Unnecessary '.ToString()' call in string concatenation
It is little bit more complicated while concatenating 'tostringed' objects only.

Example:
```C#
class C
{
    string M() => 1 + 2.ToString();
}
```

CC0118 is detected and code after fix looks like this:
```C#
class C
{
    string M() => 1 + 2;
}
```
which is incorrect.

Solution 1: Do not detect CC0118 in such situations.
Solution 2: Fix needs to be smarter.

This is problematic primarily when using solution-wide fix because of possibly large number of wrong fixes.

MaStr11:
Fixes #866.

I implemented [option 2](https://github.com/code-cracker/code-cracker/issues/866#issuecomment-309614901) as requested. This fix now looks up the types of both sides of the plus operator and tries removes those candidates, that might break things.

It turned out, that some of the checks in [`CheckAddOperationOverloadsOfTypes`](https://github.com/MaStr11/code-cracker/blob/CC0118RemoveToStringOnlyIfItIsSaveToDo/src/CSharp/CodeCracker/Style/UnnecessaryToStringInStringConcatenationAnalyzer.cs#L116) are not needed (After the second check every other check returns `false`). Essentially right now the removal of "ToString" is only save if either the underlying type is a string (as in `1 + "a".ToString()`) or the type on the other side of the plus operator is a string (as in "a" + 1.ToString()). I kept all the other checks for future enhancements to make clear that there are several cases to consider. Further I added lots of unit tests to cover these cases. The idea is that `CheckAddOperationOverloadsOfTypes` can easily be expanded in the future to support more cases without breaking anything. Those redundant checks are also very cheap and are essentially just simple comparisons.

I also had some troubles with the merging (started of the master branch, tried to re-base my changes on the V1.0.x branch, discovered that you removed this analyzer there, went back to master). So I'm sorry for the complicated commits in this pull request.


MaStr11:
Currently the tests fails because I have used some [C#7 features](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#pattern-matching):

`if (toStringCall.Expression is MemberAccessExpressionSyntax memberAccess)`

Should I remove those features or can you upgrade AppVeyor?

giggio:
The problem was in psake. I wrote a workaround now on 8f48dfe. More info: https://github.com/psake/psake/issues/201
The build passed now. Now I can check the PR.

PR #961, opened 2017-12-29
Fix issue/958 Don't suggest changes for Async Main

### LINKED ISSUE #958: CC0061 shouldn't pop for async Main
#Bug

In C# 7.1, you can have an async main. CC0061 suggests to rename to MainAsync, but that would lead to an unusable program.

https://blogs.msdn.microsoft.com/mazhou/2017/05/30/c-7-series-part-2-async-main/

````csharp
public static Task Main();
public static Task<int> Main();
public static Task Main(string[] args);
public static Task<int> Main(string[] args)
````

steamos:
Fixes #958


carloscds:
Hi @steamos. Can you create some unit tests for this fix ?

Issue #971, opened 2018-05-03
CC0127: Sort Usings Directives by length


BugAgain:
# New analyzer

There is only one option for `Removing and Sorting Usings` alphabetically in Visual Studio now. But it would be great if we could sort Usings by size without removing them. While this sounds like a useless thing, I believe my classes will look a lot more organized that way.

Before:

````csharp
using CodeCracker.Properties;
using Microsoft.CodeAnalysis;
using Microsoft.CodeAnalysis.CodeActions;
using Microsoft.CodeAnalysis.CodeFixes;
using Microsoft.CodeAnalysis.CSharp;
using Microsoft.CodeAnalysis.CSharp.Syntax;
using Microsoft.CodeAnalysis.Formatting;
using Microsoft.CodeAnalysis.Text;
using System.Collections.Generic;
using System.Collections.Immutable;
using System.Composition;
using System.Linq;
using System.Threading;
using System.Threading.Tasks;
using System;
using CodeCracker.FixAllProviders;
````

After:

````csharp
using System;
using System.Linq;
using System.Threading;
using System.Composition;
using CodeCracker.Properties;
using Microsoft.CodeAnalysis;
using System.Threading.Tasks;
using System.Collections.Generic;
using Microsoft.CodeAnalysis.Text;
using CodeCracker.FixAllProviders;
using System.Collections.Immutable;
using Microsoft.CodeAnalysis.CSharp;
using Microsoft.CodeAnalysis.CodeFixes;
using Microsoft.CodeAnalysis.Formatting;
using Microsoft.CodeAnalysis.CodeActions;
using Microsoft.CodeAnalysis.CSharp.Syntax;
````
If you agree, I can work on it.

-----------------

So in the future, maybe we can add more sorting options. Or maybe, static Usings suggestions when possible as well. e.g.

Before:

````csharp
using System;

namespace Foo
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Hello Word!");
            Console.ReadKey(true);
            [...]
````
After:

````csharp
using static System.Console;

namespace Foo
{
    class Program
    {
        static void Main(string[] args)
        {
            WriteLine("Hello Word!");
            ReadKey(true);
            [...]
````

What do you think?

Diagnostic Id: `CC0127`
Category: `Refactoring`
Severity: `Hidden`

BugAgain:
[...]

PR #977, opened 2018-05-20
Usings sort

BugAgain:
Fixes **CC0127: Sort Usings Directives by length** #971

This is an experimental PR so you can give it a try if you want before you decide if this will be helpful to you.

_P.S. I'm not that good at writing analyzers, I just wrote this because I wanted to use in a project that I'm working on. So, you can help me improve the code if you agree with this type of analyzer._

I wonder what you think.

Thx.

### DIFF
--- a/CodeCracker.sln
+++ b/CodeCracker.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26430.14
+VisualStudioVersion = 15.0.27130.2026
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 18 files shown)


giggio:
As mentioned on #971 let's wait to see if this proposal gathers more interest from the community.

CodeCracker.sln
```diff
@@ -154,4 +154,7 @@ Global
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {EF50C6AB-1421-41AE-8CB3-927AB24A69EA}
 	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {15B937E4-F880-478B-B387-3438FDD895F7}
+	EndGlobalSection
```
sharwell: 📝 The changes in this file appear to be a mistake. It introduces an unexpected duplicate section.

PR #997, opened 2018-09-10
Switch to PackageReference

sharwell:


carloscds:
@sharwell Hi, can you check errors on build ?

sharwell:
@carloscds Yep, I've made some progress but haven't reached the end yet

PR #1003, opened 2018-10-04
#928

### LINKED ISSUE #928: CC0017 Autoproperty should not trigger with `ref` and `out` uses
# Enhancement

The code below triggers CC0017 AutoPropertyAnalyzer on `MyThing`, even though the explicit implementation is needed since `_myThing` is used as a `ref` and an `out` variable.

Since this is a quite common occurrence, it would be great if CC0017 didn't trigger when the field is used as a `ref` or an `out` anywhere.

````csharp
class CC0017
{
    public int MyThing
    {
        get { return _myThing; }
        set { _myThing = value; }
    }
    private int _myThing;

    public void IncrementRef()
    {
        Interlocked.Increment(ref _myThing);
    }

    public static void SetOut(out int field)
    {
        field = 42;
    }
        
    public CC0017()
    {
        SetOut(out _myThing);
    }
}
````

carlilord:
Autoproperty should not trigger with `ref` and `out` uses
Fixes #928

### DIFF
--- a/src/CSharp/CodeCracker/Style/SwitchToAutoPropAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Style/SwitchToAutoPropAnalyzer.cs
@@ -3,7 +3,6 @@ using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
-using System;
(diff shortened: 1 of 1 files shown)


carloscds:
@carlilord You need to create unit tests for all changes that you made, ok!

carlilord:
will do

carloscds:
@carlilord Some news about this ?

carloscds:
@carlilord Sure!

PR #1004, opened 2018-10-06
Adding an "Async" method suffix should remove any embedded "Async" string

### LINKED ISSUE #922: Adding an "Async" method suffix should remove any embedded "Async" string in CC0061
# Enhancement

CC0061 is often triggered due to an additional suffix having been added to an `async` method name, resulting in an embedded "Async":

````csharp
public async static Task MyAsyncHelper()
{
    await Task.Delay(1);
}
````

Current output after fix applied, in v1.0.3:

````csharp
public async static Task MyAsyncHelperAsync()
{
    await Task.Delay(1);
}
````

Expected output after fix applied, where any embedded "Async" string have been removed:

````csharp
public async static Task MyHelperAsync()
{
    await Task.Delay(1);
}
````

carlilord:
Fixes #922

### DIFF
--- a/src/CSharp/CodeCracker/Performance/RemoveWhereWhenItIsPossibleAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Performance/RemoveWhereWhenItIsPossibleAnalyzer.cs
@@ -25,15 +25,7 @@ namespace CodeCracker.CSharp.Performance
             "Any",
             "Single",
             "SingleOrDefault",
-            "Count",
-            "FirstAsync",
-            "FirstOrDefaultAsync",
-            "LastAsync",
-            "LastOrDefaultAsync",
-            "AnyAsync",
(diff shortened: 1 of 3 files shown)


carloscds:
@carlilord I guess that you can simplify your code, removind HasAsyncCorrectlyInName() method. Principal goal is put "Async" name at end of method, and to fix this, you need remove "Async" with from name and put at the end. What do you think about?

carlilord:
I did that because in the ticket it is mentioned, that some words could contain "async" itself. So I added this check additional checks in the function

carloscds:
@carlilord You are right. Please now, add unit tests! and for all changes on this project, please put unit tests 👍

carlilord:
Finally have time to continue on these issues. But for this PR there are already unit tests checked in.

PR #1005, opened 2018-10-08
Make CC0011 work with IQueryable and Async methods

### LINKED ISSUE #905: Make CC0011 work with IQueryable and Async methods
This is not a bug, rather aIf  very helpful extension to the [CC0011 rule](http://code-cracker.github.io/diagnostics/CC0011.html) to also skip `Where` conditions on `IQueryable` extension methods and when using for instance `FirstOrDefaultAsync()`

Example with EF Core code:

````csharp
await _context.MyTable.Where(t => t.Col1 == 1).FirstOrDefaultAsync();
````

Expected output after change applied:

````csharp
await _context.MyTable.FirstOrDefaultAsync(t => t.Col1 == 1);
````

If I read the rule correctly, than it should be enough to add `Async` suffix to the strings defined here: https://github.com/giggio/code-cracker/blob/master/src/CSharp/CodeCracker/Performance/RemoveWhereWhenItIsPossibleAnalyzer.cs#L20

Here are the latest EF-Core extension methods: https://github.com/aspnet/EntityFramework/blob/8bac118ee8b7ff5c30918e95682f454e0233c23e/src/EFCore/EntityFrameworkQueryableExtensions.cs - They seem to define all the async counterparts of the methods already used in the CC0011 rule

carlilord:
Added async extension versions to the analyzer
Fixes #905

### DIFF
--- a/src/CSharp/CodeCracker/Performance/RemoveWhereWhenItIsPossibleAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Performance/RemoveWhereWhenItIsPossibleAnalyzer.cs
@@ -25,8 +25,17 @@ namespace CodeCracker.CSharp.Performance
             "Any",
             "Single",
             "SingleOrDefault",
-            "Count"
+            "Count",
+            "FirstAsync",
+            "FirstOrDefaultAsync",
+            "LastAsync",
+            "LastOrDefaultAsync",
+            "AnyAsync",
+            "SingleAsync",
+            "SingleOrDefaultAsync",
+            "CountAsync"
         };
+
         internal static readonly DiagnosticDescriptor Rule = new DiagnosticDescriptor(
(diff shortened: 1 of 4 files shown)


carloscds:
@carlilord Please, create unit tests for your changes!

carlilord:
For the creation of tests I would need to add the Microsoft.EntityFrameworkCore package, as this package contains the wanted Extension methods but I get the following error:

`
Could not install package 'Microsoft.EntityFrameworkCore 2.1.4'. You are trying to install this package into a project that targets '.NETFramework,Version=v4.5.2', but the package does not contain any assembly references or content files that are compatible with that framework. For more information, contact the package author.
`

Should I adapt the project structure, or should I just create dummy extensions?

carloscds:
@carlilord You need to create without EF package. You can use something like this:
IQueryable<Customer> list = new List<Customer>().AsQueryable();

Issue #543, opened 2015-10-09
Is there any way to config inspect severity? Some errors are not necessary but noisy


iccfish:
I wonder the possibility of add an inspect severity configuration like R#, due to some analysis errors are not really need to be fixed.

For example, checkout the source codes of [Dapper.Net](https://github.com/StackExchange/dapper-dot-net)  and open it in visual studio, almost 200+ errors were shown, and most of them are ```CC0090 incorrect xml documention```, ```CC0021 use nameof```.
![dapper_error](https://cloud.githubusercontent.com/assets/1091653/10391778/db08a428-6eb3-11e5-8bfb-f792b3a5f877.png)

If I use ```#pragma warning disable CC0090``` to disable CC0090, there are still 170+ errors, and says ```CC0001 Use var ``` ...

![dapper_error_2](https://cloud.githubusercontent.com/assets/1091653/10392039/1fe4fae0-6eb6-11e5-9f8d-f9f2c91f019b.png)

Of cause these are issues, but they don't need to be changed or rewrote due to they are third party projects or they are old projects, and works well currently. And I don't want make any changes to these projects even a little modification.
So it will be more effective if inspect severity could be configured.

Btw, I want to contribute on translation. I've already read the contributing guideline, and knew I need to start a new issue. But I don't know where to find all diagnostics (seems that [this page](http://code-cracker.github.io/diagnostics.html) was out-dated) and what's the content format of the issue.

carloscds:
@iccfish We don't have documentation for all analyzers at this time, but you can see IDs at this page: https://github.com/code-cracker/code-cracker/wiki/DiagnosticIds

giggio:
@iccfish You can already change the severity or even disable specific diagnostics. See https://github.com/code-cracker/code-cracker/issues/47#issuecomment-63705397 for details on how to do that. This is a Visual Studio/.NET thing, we just benefit from it. :)
I will be closing the issue, as it is already possible to do what you need. If you find there is more we can do, let us know.

giggio:
@iccfish Regarding translation, not all analyzers are yet ready for translation. We would love to have translation happening. The infrastructure is the same as from any localized .NET assembly, based on resources.
But we still need to prepare the infrastructure for all the analyzers and fixers. We have issue #360 tracking that, with an example on how to do it. We would love your collaboration either making the analyzers and fixes localizable, or on the translation itself. What language are you interested in?

iccfish:
[...]

Issue #524, opened 2015-09-28
Can't Manage extension settings


ImaginaryDevelopment:
I've installed Code Cracker as an extension for Visual Studio 2015, but there appears to be no way to manage which diagnostics are on and what levels they are on at. I've found the analyzers node of References in solution explorer. While I can make changes there they don't seem to save as the new set the next time I open the solution or project. Even if they did, this is installed for my entire VS2015, so I'd like to manage it at that level rather than have to make changes to every project I open.

giggio:
You can do it in VS. See this comment: https://github.com/code-cracker/code-cracker/issues/47#issuecomment-63705397

ImaginaryDevelopment:
The link provides a way to set it per project, meaning for every project I open I have to go set the ruleset.

giggio:
You can share the ruleset file with all the projects. Just move it to the root folder where the solution lives. This is what we are doing at CodeCracker, check the repo.

PR #1036, opened 2019-08-30
Make all diagnostic severities configurable

### LINKED ISSUE #47: Allow user to download custom combination of analyzers and refactorings
**This is a low priority issue, if you want to work on it please read the comments**

CodeCracker is bringing together tens of different analyzers and refactorings and very soon it will gather hundreds. Not everyone will want to use all of them. This issue focuses on allowing the user to choose whatever combination of analyzers and refactorings they want and generate a custom package. A plus would be to publish that to nuget so they can use that directly on their project without the need to download, and if a combination already exists point them to an existing package.

We could build something like the [Jquery UI download builder](http://jqueryui.com/download/).

We don't have any idea on how to do that yet, so we need to discuss it. We could probably work with `csc` or `msbuild` directly to build an exact package combination. We would also need to work on the isolation of the analyzers and refactorings, so they can be used without depending on each others. And this all needs to be testable.

This issue is made harder by the fact that we do not yet have a build server.

Ideas are welcome and needed. This issue is first an invitation to discussion, and, after we decide on a strategy, it will become the actual implementation of the chosen strategy. So, please, brainstorm!

walterlv:
Fixes #47 , #524 , #543  .

### DIFF
--- a/src/CSharp/CodeCracker/Design/CatchEmptyAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Design/CatchEmptyAnalyzer.cs
@@ -18,7 +18,7 @@ namespace CodeCracker.CSharp.Design
             Title,
             MessageFormat,
             Category,
-            DiagnosticSeverity.Warning,
+            SeverityConfigurations.Current[DiagnosticId.CatchEmpty],
             isEnabledByDefault: true,
             helpLinkUri: HelpLink.ForDiagnostic(DiagnosticId.CatchEmpty));
 
--- a/src/CSharp/CodeCracker/Design/EmptyCatchBlockAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Design/EmptyCatchBlockAnalyzer.cs
@@ -20,7 +20,7 @@ namespace CodeCracker.CSharp.Design
             Title,
             MessageFormat,
             Category,
-            DiagnosticSeverity.Warning,
+            SeverityConfigurations.Current[DiagnosticId.EmptyCatchBlock],
             isEnabledByDefault: true,
             description: Description,
             helpLinkUri: HelpLink.ForDiagnostic(DiagnosticId.EmptyCatchBlock));
--- a/src/CSharp/CodeCracker/Design/MakeMethodStaticAnalyzer.cs
+++ b/src/CSharp/CodeCracker/Design/MakeMethodStaticAnalyzer.cs
@@ -1,10 +1,11 @@
-﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using System.Linq;
 using System.Collections.Immutable;
 using System;
+using CodeCracker;
 
 namespace CodeCracker.CSharp.Design
 {
@@ -23,7 +24,7 @@ namespace CodeCracker.CSharp.Design
             Title,
             MessageFormat,
             Category,
-            DiagnosticSeverity.Warning,
+            SeverityConfigurations.Current[DiagnosticId.MakeMethodStatic],
             true,
             description: Description,
             helpLinkUri: HelpLink.ForDiagnostic(DiagnosticId.MakeMethodStatic));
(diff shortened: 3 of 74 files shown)


carloscds:
@walterlv Can you explain benefits of this big change ?

walterlv:
@carloscds I've extracted all the Diagnostic Severities into one configuration class in this pull request. Then we can load a custom ruleset file later to use the user's preference. This makes the custom diagnostic severities work well as a Visual Studio extension not only as NuGet packages.

sharwell:
I'm super confused. Unless you are using `WellKnownDiagnosticTags.NotConfigurable` (which the analyzers here do not appear to be doing), the severities are already configurable without this change.

sharwell:
@carloscds I don't think this change is needed. If you reach a point where you do not agree with this, please reach out to me so we can make sure it's not just a failure on our part to document the analyzer behaviors. 😄

walterlv:
@sharwell As a NuGet package, the code-cracker works well and the MSBuild automatically use the user's ruleset file to configure our diagnostic severities. But as a Visual Studio extension, the MSBuild does nothing for these analyzers so we need to configure the severities ourselves. I change these codes only for the configuration when we use it as a Visual Studio extension.

carloscds:
@sharwell I agree with you, code is very complex after this change.

sharwell:
@walterlv After this change, exactly how would one configure the severity inside Visual Studio? Currently ruleset files work with VSIX-installed analyzers, and starting with 16.3 you can use .editorconfig to set the severity for VSIX-installed analyzers. Neither of these approaches requires changes to the implementation code.

walterlv:
@sharwell you still don't get the main reason why I should do this! How can I add .EditorConfig or ruleset for all projects? Some of them are even not mine.

I can't add a code-cracker only ruleset for Microsoft GitHub projects, or for other team's projects. So there are actually no available custom ruleset for vsix-installed analyzers.

If I can’t configure the securities globally, I have to ruin all projects with my custom ruleset files. Then no pull requests will be merged in any team's GitHub projects because no one will accept a private preference file that they don't even know what it is!

walterlv:
By the way, I can't find the description that the global ruleset will work in 16.3 preview.

[Visual Studio 2019 Preview Release Notes](https://docs.microsoft.com/en-us/visualstudio/releases/2019/release-notes-preview)

walterlv:
I should add the ruleset file loading code for this pull request so that the user can change the securities globally.

walterlv:
@sharwell The core project is a portable project so I can't generate a configuration file path. How can I read the custom user's ruleset in codes only for this analyzer extension if I can't find the configuration path?

sharwell:
@walterlv From your explanation, it makes sense why you can't use .editorconfig or ruleset files. The part which is still not clear to me is which alternative you are planning to use if this is merged.

walterlv:
@sharwell Thanks for your reply. From the description of this pull request, you can find that I plan to read a custom ruleset file from file system but I don't know how to do it in a portable project. Could you give me some suggestions?

jwooley:
@sharwell I see that 16.3 includes support for configuring the rules via .editorconfig. Is there design documentation available on how this works that are public? If it can pick up an .editorconfig from the solution root, when installing the tools as a vsix? would that solve the issue that @walterlv  is trying to accomplish. I would be hesitant to make this large of a change if the tooling is moving towards embracing .editorconfig more fully.

sharwell:
> Is there design documentation available on how this works that are public?

https://github.com/dotnet/roslyn/blob/master/docs/compilers/analyzer-config.md

> If it can pick up an .editorconfig from the solution root, when installing the tools as a vsix?

Yes, it would pick them up like this in 16.3.

walterlv:
@jwooley I've posted an opinion above that .editorconfig is not a right solution:

> If I can’t configure the severities globally, I have to ruin all projects with my custom ruleset files. Then no pull requests will be merged in any team's GitHub projects because no one will accept a private preference file that they don't even know what it is!

This means that I can't put an .editorconfig file on every solution repositories.

sharwell:
@walterlv you can place it in the parent folder of the solution (outside/above the source control directory) and it will still get picked up.

walterlv:
@sharwell Thanks.

I've tried multiple .editorconfig files both inside and outside the git repository. I find that both of them work. So I can prepare a nearly global .editorconfig file to my drive partition root directory.

This helps me instead of config it inside a settings UI like below:

image

walterlv:
I've recommended Code Cracker via my blog post:

- <https://blog.walterlv.com/post/recommend-some-code-analysis-extensions-or-nuget-packages.html>

Issue #903, opened 2017-03-21
BUG: CC0052 - ReadonlyFieldAnalyzer duplicates preprocessor directive


nishant95:
#Bug
The code-fix of CC0052 (ReadonlyFieldAnalyzer) duplicates the  ```` #region ````

Before:

````csharp
public class DemoController : Controller
{
        #region Demo region

        IDemoService _demoService;
        IMapper _mapper;

        #endregion
}
````

Current output after code-fix applied:

````csharp
public class DemoController : Controller
    {
        #region Demo region

        readonly
        #region Demo region

        IDemoService _demoService;
        IMapper _mapper;

        #endregion
}
````

Expected output after fix applied:

````csharp
public class DemoController : Controller
    {
        #region Demo region

        readonly IDemoService _demoService;
        IMapper _mapper;

        #endregion
}
````

bzier:
Same behavior here. Can confirm using the extension version 1.0.3 in VS 2017.

giggio:
Bug confirmed. I made the description cleaner to make the bug clearer.

CygnusSW:
If this one is still available, I can grab it.

carloscds:
@callenWrench Go ahead.

PR #1038, opened 2019-10-05
#903 (#CC0052): Fix for Duplication of Directives When Fixing Readonly

cj-the-affable:
Fixes # 903.
This PR is fixed the duplication of regions created when the CC0052 diagnostic fix is applied to one or more variable declarations.


callenWrench:
Putting this on hold for a little while. Fixing the end-of-line discrepancy and cleaning up the commit history.

DanielSantaRosaZup (approved):

